PR #7, opened 2018-11-02
feat: add support for Storybook, run as 'yarn storybook'

riccardo-forina:


.storybook/webpack.config.js
```diff
@@ -0,0 +1,35 @@
+const path = require("path");
+
+module.exports = (baseConfig, env, defaultConfig) => {
+  // Extend defaultConfig as you need.
+
+  // For example, add typescript loader:
+  defaultConfig.module.rules.push({
+    test: /\.(js|mjs|jsx|ts|tsx)$/,
```
seanforyou23: Is jsx support required or desirable for some reason? I like the idea of shoring up the file types, if we'll benefit from denoting js modules with .mjs or using jsx file extension then let's make a recommendation around when to use them. Otherwise, can we just stick to js, ts, and tsx file types?

riccardo-forina: To be fair, I copy-pasted those scripts from the official guide. I guess we can tune this down a notch to `ts|tsx`files only.

PR #12, opened 2018-11-06
Init cypress testing framework

seanforyou23:
This PR stands up cypress for helping with our testing efforts for the first time. There's currently one test that goes through the form on the settings page, and verifies that each label is properly bound to its form control counterpart, something that's been a problem on and off for us in the past for the settings page oauth form.

A lot of this changeset is adding initial files, the interesting changes are called out in the paths below.

Remove conflicting `@types/jest` package
Move example tests outside of active test directory
Fix a11y issue on `src/app/SettingsPage.tsx`
First test at `cypress/integration/components/settings.spec.js`


seanforyou23:
@riccardo-forina the existing test script for CRA never worked for me even on master branch. I gave up trying to fix it. Figured whatever it was attempting to do would be covered by future work toward "testing framework investigation" anyway.

PR #13, opened 2018-11-06
feat(a11y): let us keep accessibility compliance in check

seanforyou23:
This PR adds a simple yet effective means to keeping an eye on accessibility support within the UI. React Axe is a package ([used in Patternfly core](https://github.com/patternfly/patternfly-next/issues/466)) that runs on page load and analyzes the page's markup for common issues found in WCAG and similar specifications. Feedback is supplied in the browser's developer console as seen below.

A great way to keep a pulse on a11y as we're starting fresh with new a UI platform, developers don't need to do anything special - just notice the feedback when supplied in the developer console. If there are no issues found, the tool will not report anything.

screen shot 2018-11-06 at 9 00 08 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "@types/node": "^10.11.3",
     "@types/react": "^16.4.14",
     "@types/react-dom": "^16.0.8",
+    "react-axe": "^3.0.2",
     "typescript": "^3.1.1"
   },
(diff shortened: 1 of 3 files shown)


seanforyou23:
Closing this for now, simple enough to resume later once we get back to the testing bits.

Issue #15, opened 2018-11-13
Incremental Docs Needed


seanforyou23:
As a potential contributor, I need help with way-finding to get started with contributing to this POC work. Currently, our README is pretty empty in that it only contains a random snippet of code, not explaining what it does or when it needs to be run.

Obviously the nature of POC work is subject to frequent change and so documenting certain things just doesn't make sense until they are further solidified. That said, surely there is some level of detail that can be provided so users can stand up the development environment in the basic sense.

I think documenting the basic steps required to stand the app up in "hello world" fashion would be good to do, particularly around how to handle authentication. Otherwise requests to `http://syndesis-server-syndesis.xxx.xxx.xx.x.nip.io/api/v1/` fail and the initial developer experience isn't so fruitful.

riccardo-forina:
I wrote a README documenting how the project is organized [here](https://github.com/riccardo-forina/syndesis-react-poc/blob/multi-app/README.md), and how to make the first setup [here](https://github.com/riccardo-forina/syndesis-react-poc/blob/multi-app/README.md#first-time-setup).

Hope it helps.

gashcrumb:
I think the readme probably satisfies this, unless @seanforyou23 there's any outstanding items we should add?

gashcrumb:
@elvisisking if you can think of any other first-timer doc improvements feel free to chime in here too :-)

seanforyou23:
Those README updates were exactly what I was hoping for. LGTM - I'll go ahead and close this, @elvisisking feel free to reopen or create a new issue if you find the need.

Issue #11, opened 2018-11-05
Components should be model agnostic


riccardo-forina:
Components should stick to basic types and refrain to depend on application models, like [here](https://github.com/syndesisio/syndesis-react-poc/blob/master/src/components/Connection.tsx#L6), to ease testing and reduce coupling.

riccardo-forina:
Done in #16

PR #16, opened 2018-11-14
Reorganize the project and document it

riccardo-forina:
Should we promote this to be the new master branch?

This fixes some open issues, like #11 and #15.


lerna.json
```diff
@@ -0,0 +1,10 @@
+{
+  "lerna": "3.4.3",
+  "packages": [
+    "syndesis",
+    "packages/*"
+  ],
+  "npmClient": "yarn",
+  "useWorkspaces": true,
+  "version": "0.1.0"
+}
```
kahboom: This drives me a bit nuts tbh.. 🙈 such a nitpick on my part though

packages/api/.storybook/addons.js
```diff
@@ -0,0 +1 @@
+import "storybook-addon-specifications/register";
```
kahboom: Newline here too, please?

packages/api/.storybook/config.js
```diff
@@ -0,0 +1,21 @@
...
+  req.keys().forEach(filename => req(filename));
+}
+
+addDecorator(withInfo({
+  inline: true,
+  header: false,
+  maxPropsIntoLine: 1
+}));
+configureEnzyme({ adapter: new Adapter() });
+configure(loadStories, module);
```
kahboom: and here!

packages/api/src/ApiContext.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
```
kahboom: Noticed some `import`s are single quote and others double--personally I prefer single, but I don't mind as long as we're consistent I think.

packages/api/src/SyndesisRest.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import { ApiContext } from "./ApiContext";
```
kahboom: Mixing single and double quotes here

packages/connector-ftp/.storybook/StoryHelper.css
```diff
@@ -0,0 +1,19 @@
...
+@import url('~patternfly/dist/css/patternfly-additions.css');
+@import url('~patternfly-react/dist/css/patternfly-react.css');
+
+html,
+body,
+#root {
+  height: 100%;
+}
+
+#root {
```
kahboom: Not too familiar with Storybook or this setup, but is it necessary to use an ID over a class here?

kahboom:
:D ty!

README.md
```diff
@@ -1,12 +1,267 @@
...
+
+Or
+
+```bash
+# From the syndesis folder
+$ yarn start
+```
+
+
+#### First time setup
```
seanforyou23: Might be good to move this section before the "Development Server" section, so the steps are more sequential.

README.md
```diff
@@ -1,12 +1,267 @@
...
+  to:
+    kind: Service
+    name: syndesis-server
+    weight: 100
+  wildcardPolicy: None
+')
+```
+
+Now we can setup the app. Run the [syndesis development server](#development-server) and open the address in a browser.  
+You will be asked to provide two information:
```
seanforyou23: This suggests that I'd be redirected to the settings page upon navigating to the dev server url. I remember seeing this happen before, but now for some reason I don't get automatically redirected to settings and instead the UI partially loads whatever view I'm navigated with the the network requests to connections/integrations failing.

Might be helpful to just mention to navigate to the settings page to enter this information unless we can have the redirects work again?

README.md
```diff
@@ -1,12 +1,267 @@
...
+ - "http://localhost:5000/"
 grantMethod: prompt
 ')
-```
\ No newline at end of file
+```
+
+Finally, expose the server with a direct route.   
+**Replace the `CLUSTER_ADDRESS` placeholder with the right address of your cluster.**
+For a local development environment running on Minishift, it will look like this: `192.168.64.1.nip.io`
```
seanforyou23: I made a mistake and left off the `.nip.io` extension the first time around. When I re-ran the same command, just with the proper url this time, I get the following error:

`Error from server (AlreadyExists): error when creating "/dev/fd/63": routes.route.openshift.io "syndesis-server" already exists`.

I imagine this may happen to others as well, can we inform how to see all available routes and how to delete the old route we've exposed?

gashcrumb: I think shortly after merging this we'll automate this stuff via a script to be honest, as having to run these commands by hand is error prone.

README.md
```diff
@@ -1,12 +1,267 @@
...
+```bash
 oc create -f <(echo '
 kind: OAuthClient
 apiVersion: oauth.openshift.io/v1
 metadata:
  name: camel-k-ui
 secret: "..."
 redirectURIs:
- - "http://localhost:3000"
+ - "http://localhost:5000/"
```
gashcrumb: Sure it shouldn't be 3000?  The dev server for me came up at that port.

README.md
```diff
@@ -1,12 +1,267 @@
...
+$ oc login -u admin
+```
+
+Then, create a new OAuth Client:
+```bash
 oc create -f <(echo '
 kind: OAuthClient
 apiVersion: oauth.openshift.io/v1
 metadata:
  name: camel-k-ui
```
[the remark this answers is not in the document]
gashcrumb: The oauth client uses this as the client ID, so it matters.  Eventually we'll probably change this to "Syndesis" or something, just that currently using a different oauth client entirely means that you can also still run the old console code still against the same minishift instance.

README.md
gashcrumb: I'm thinking that eventually we'll want to adjust this to use a `config.json` file, which would then be used to specify the API endpoint and oauth endpoints.  The main reason is we already have a bit of infrastructure in place to support this method; the `config.json` file is loaded in the real app via a config map, we could probably use the same approach for these settings.

README.md
riccardo-forina: Good catch! It should totally be 3000

riccardo-forina: Good catch! It should totally be 3000

README.md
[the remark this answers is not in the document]
riccardo-forina: Ah, it’s a leftover from the camel-k poc. As Stan said it matteres but we can change it to whatever.

README.md
```diff
@@ -1,12 +1,267 @@
...
 secret: "..."
```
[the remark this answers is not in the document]
riccardo-forina: Well it’s a secret, so in the readme is always going to be some sort of placeholder. But since we use the implicit grant flow the secret doesn’t really matter.

README.md
riccardo-forina: @seanforyou23: the redirect happens only when the settings are empty, so what you experienced is right.

But yeah, the “app” component design is totally in a POC state. As Stan said, we should “talk” with Syndesis using the constructs that are already available, but for the moment I think we can continue with what we have.

riccardo-forina:
Ah! I see. Yeah the problem is that I didn’t consider the initial “nothing has been built yet” state when running `yarn watch`. The fix is easy, instead of running the two `watch:app` and `watch:packages` with `npm-run-all` I should run all the watchers using `lerna` directly, that will take care of running packages taking the dependency tree in consideration.

Or something like that 😂

I’ll check it out tomorrow.

README.md
```diff
@@ -1,12 +1,267 @@
...
+
+```bash
+# From the repository root
+$ yarn watch:app
+```
+
+Or
+
+```bash
+# From the syndesis folder
```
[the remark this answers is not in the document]
riccardo-forina: Not really, it's just a hint to how Lerna works. I'd keep it this way for the moment.

README.md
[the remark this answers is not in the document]
riccardo-forina: Done

riccardo-forina:
Ok I pushed an update that:

* should fix the build from a clean slate
* improve the documentation
* brings Jest to all the packages
* improves the Storybook for the `@syndesis/ui` module

_Maybe_ we are almost ready to start working 😂

riccardo-forina:
This is how I set up the Storybook. @seanforyou23 I added the a11y plugin you linked the other day 👍 

storybook

gashcrumb:
I'm +1 for merging personally...

README.md
seanforyou23: Ok makes sense, thx for the heads up!

README.md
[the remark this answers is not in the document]
seanforyou23: Ok I see. I suppose it's just worth noting if this value will be needed for anything in the future or if you can just enter whatever. As a first timer, I'm wondering if I need to keep up with the value I enter here, and if so, for what reason.

README.md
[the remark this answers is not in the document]
seanforyou23: Thx for explaining! Makes sense now.

seanforyou23:
Awesome work @riccardo-forina! This really takes a huge step forward.

riccardo-forina:
🎉

PR #28, opened 2018-11-20
Load Atlasmap in Syndesis using an iframe

### LINKED ISSUE #27: Research loading a 3rd party component via iframe
Let's first try and load a generic simple angular 6 app, and then see if we can load atlasmap into an iframe easily enough.  We should also define the communications protocol between the host (syndesis) and the component.

There's a bunch more things we talked about here that I can't remember exactly too.

riccardo-forina:
### TestAtlasmapPage.tsx

This shows how to use an iframe with the srcDoc attribute with the injection of a dynamically loaded module (Atlasmap in this case).

Todo:
* [x] extract the iframe to its own UI component
* [x] define a communication protocol to pass data between the host and the document and vice versa (postMessage)
* [x] ~kill this file~ will keep for a while as a mean to test the iframe

## packages/atlasmap (aka @syndesis/atlasmap)

An Angular 7 app that pulls Atlasmap as a dependency and exposes the Data Mapper as a stand-alone app. Data Mapper assets must be copied in the public/atlasmap folder of the syndesis app (this is already done automatically by `yarn build` in the root).

Todo:
* [x] make the app configurable from the outside, using the communication protocol defined above
* [x] this now sits in the extra folder that's not handled by Lerna because of some initial problem with clashing dependencies between this and the rest of the app. It would be better to have this inside the packages folder, to reduce build complexity (although right now everything should just work with a `yarn build` from the root).

## Preview

atlasmap-iframe

Fixes #27


gashcrumb:
So this works really well here for me...  When are we ready to merge?

riccardo-forina:
Whenever you feel it :D

gashcrumb:
Perfect!  One rebase away from being merged I think...

PR #29, opened 2018-11-26
feat: fetch when notified by the server side event service

### LINKED ISSUE #26: Use event service to fetch updates on-demand instead of polling
Right now the app polls but we do have a websocket that we could connect to that sends object change notifications that could be used so that we only fetch from the server when needed.

gashcrumb:
fixes #26 

Currently still just a POC, basically I'd like to check if I'm going in the right direction with this @riccardo-forina 

The general idea is to connect to the server side event source once and make the message queue available to child components so they can listen and act on incoming changes as needed.


riccardo-forina:
Just to be clear, the code I wrote in the above message is what I wanted to when refactoring the api package to clean it up. Right now the `Rest` component has too many concerns (like the polling, that should be extracted to its own thing).

Also, when hooks will land converting those components to hook should be pretty straightforward.

gashcrumb:
@riccardo-forina okay, overhauled it a bit :-)

I pulled out the polling from the Rest component there and created a WithPolling component.  In the main app the WithServerEvents component gets created which initializes the one instance of the listener for server events, then a context is used to hook up the WithChangeListener component so it can receive the change events.  So far in this PR I just hooked up the integrations page, I added the poller to the connection and connector lists but will probably swap 'em out tomorrow and then also deal with the dashboard.  After incorporating any feedback you have of course, thanks!

packages/api/src/CallFetch.tsx
```diff
@@ -0,0 +1,34 @@
+export interface IHeaders {
```
riccardo-forina: I think the filename should start with a lowercase, and its extension be just `ts` since there is no JSX inside. So `callFetch.ts`

riccardo-forina: Maybe since we are here, also rename `IHeaders` to `IFetchHeaders`?

packages/api/src/Rest.tsx
```diff
@@ -56,21 +23,17 @@ export interface IRestState<T> {
 export interface IRestProps<T> {
   autoload?: boolean;
   baseUrl: string;
-  poll?: number;
   url: string;
-  headers?: IHeader;
+  kind?: string;
```
riccardo-forina: this can be removed now

packages/api/src/Rest.tsx
```diff
@@ -79,37 +42,21 @@ export class Rest<T> extends React.Component<IRestProps<T>, IRestState<T>> {
       error: false,
       hasData: false,
       loading: true,
-      read: this.read,
-      save: this.onSave,
+      read: this.read.bind(this),
```
riccardo-forina: It's better to do the binding once in the constructor for all, so we are guaranteed that `this` is always going to point to the right thing. So
```
this.read = this.read.bind(this);
this.save = this.save.bind(this);
```

Also, why are you linking the `save` and `read` functions in the state? This will mess up with the change detection done in the `shouldComponentUpdate` hook I think.

If this is to pass them in the render function, just recompose the object on the fly, like: 

```tsx
public render() {
  return this.props.children({
    ...this.state,
    read: this.read,
    save: this.save
  );
}	   
```

packages/api/src/WithChangeListener.tsx
```diff
@@ -0,0 +1,44 @@
...
+export class WithChangeListener extends React.Component<
+  IWithChangeListenerProps
+> {
+  public constructor(props: IWithChangeListenerProps) {
+    super(props);
+    this.changeListener = this.changeListener.bind(this);
+  }
+
+  public render() {
+    return this.props.children(this.state);
```
riccardo-forina: `this.state` will always be undefined here, no need to pass it.

packages/api/src/WithPolling.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  public async componentDidMount() {
+    this.startPolling();
+  }
+
+  public async componentWillUnmount() {
+    this.stopPolling();
+  }
+  public render() {
+    return this.props.children(this.state);
```
riccardo-forina: `this.state` will always be undefined here, no need to pass it.

packages/api/src/WithServerEvents.tsx
```diff
@@ -0,0 +1,241 @@
+/* tslint:disable */
```
riccardo-forina: Just remember to remove this eventually :)

packages/api/src/WithServerEvents.tsx
```diff
@@ -0,0 +1,241 @@
...
+  private messageListeners = {};
+
+  public constructor(props: IWithEventsProps) {
+    super(props);
+    this.state = {
+      registerChangeListener: this.registerChangeListener.bind(this),
+      registerMessageListener: this.registerChangeListener.bind(this),
+      unregisterChangeListener: this.unregisterChangeListener.bind(this),
+      unregisterMessageListener: this.unregisterMessageListener.bind(this),
+    };
```
riccardo-forina: As above, bind the functions in the constructor body and don't set them in the state unless absolutely necessary.

packages/api/src/WithServerEvents.tsx
```diff
@@ -0,0 +1,241 @@
...
+    delete this.messageListeners[key];
+  }
+
+  public async componentDidMount() {
+    this.apiEndpoint = this.props.apiUri;
+    this.headers = this.props.headers;
+    this.start();
+  }
+  public render() {
+    return this.props.children(this.state);
```
riccardo-forina:
```tsx
return this.props.children({
  registerChangeListener: this.registerChangeListener,
  registerMessageListener: this.registerChangeListener,
  unregisterChangeListener: this.unregisterChangeListener,
  unregisterMessageListener: this.unregisterMessageListener,
});
```

packages/api/src/WithServerEvents.tsx
```diff
@@ -0,0 +1,241 @@
...
+    delete this.changeListeners[key];
+  }
+
+  public unregisterMessageListener(key: string) {
+    delete this.messageListeners[key];
+  }
+
+  public async componentDidMount() {
+    this.apiEndpoint = this.props.apiUri;
+    this.headers = this.props.headers;
```
riccardo-forina: Here you are just duplicating the props without doing any change. Props should be considered a single source of truth, storing their value in some internal state is a common pattern for bugs. Just refer to the `this.props.propName` value when you need it.

packages/api/src/WithServerEvents.tsx
```diff
@@ -0,0 +1,241 @@
...
+    }
+    this.starting = true;
+    try {
+      const response = await fetch(`${this.apiEndpoint}/event/reservations`, {
+        body: '',
+        method: 'POST',
+        headers: {
+          contentType: 'application/json; charset=utf-8',
+          accept: 'application/json,text/plain,*/*',
+          ...this.headers,
```
riccardo-forina: `this.props.headers`

packages/api/src/WithServerEvents.tsx
```diff
@@ -0,0 +1,241 @@
...
+          break;
+        default:
+          // Keep flipping between WS and ES untill we find one that works.
+          this.startConnection(this.retries % 2 === 0);
+          break;
+      }
+    }, reconnectIn);
+  }
+
+  private async startConnection(connectUsingWebSockets: boolean) {
```
riccardo-forina: If you don't like writing a bunch of `this.props.apiUri` you can set a `const {apiUri: apiEndpoint} = this.props;` at the top of the function.

But I found this to lessen the IDE refactoring capabilities when I actually want to rename the prop because the name I picked sucks, since I end up getting something like `const { newCoolName: oldBadName} = this.props`.

packages/api/src/WithServerEvents.tsx
```diff
@@ -0,0 +1,241 @@
...
+    if (this.starting) {
+      return;
+    }
+    this.starting = true;
+    try {
+      const response = await fetch(`${this.apiEndpoint}/event/reservations`, {
```
riccardo-forina: ~`this.apiEndpoint`~, see previous comment

packages/api/src/WithServerEvents.tsx
```diff
@@ -0,0 +1,241 @@
...
+        headers: {
+          contentType: 'application/json; charset=utf-8',
+          accept: 'application/json,text/plain,*/*',
+          ...this.headers,
+        },
+      });
+      let reservation = await response.json();
+      reservation = reservation.data;
+      if (connectUsingWebSockets) {
+        let wsApiEndpoint = _resolve(window.location.href, this.apiEndpoint!);
```
riccardo-forina: ~`this.apiEndpoint`~, see previous comment

packages/api/src/WithServerEvents.tsx
```diff
@@ -0,0 +1,241 @@
...
+          console.log('Connecting using web socket');
+        this.starting = false;
+        this.started = true;
+        return this.connectWebSocket(wsApiEndpoint);
+      } else {
+        this.starting = false;
+        this.started = true;
+        console.log('Connecting using server side events');
+        return this.connectEventSource(
+          this.apiEndpoint + '/event/streams/' + reservation
```
riccardo-forina: ~`this.apiEndpoint`~, see previous comment

syndesis/src/modules/connections/pages/ConnectionsPage.tsx
```diff
@@ -73,59 +73,69 @@ export default class ConnectionsPage extends ListViewToolbarAbstractComponent<
...
-                  loading={!hasData}
-                  loaderChildren={
-                    <>
-                      {new Array(5).fill(0).map((_, index) => (
-                        <ConnectionsGridCell key={index}>
-                          <ConnectionSkeleton />
-                        </ConnectionsGridCell>
-                      ))}
-                    </>
+            <WithPolling read={read} polling={5000}>
```
riccardo-forina: 👍 

I think we could add a `WithUpdatingConnections` helper class that hides this away from the page and keeps the logic in the `@syndesis/api` package.

syndesis/src/modules/integrations/pages/IntegrationsPage.tsx
```diff
@@ -120,91 +126,111 @@ export default class IntegrationsPage extends ListViewToolbarAbstractComponent<
...
+        {({ read, data: integrationsData, hasData, error }) => (
+          <ServerEventsContext.Consumer>
+            {({ registerChangeListener, unregisterChangeListener }) => (
+              <WithChangeListener
+                listenerKey="integration-page"
+                read={read}
+                registerChangeListener={registerChangeListener}
+                unregisterChangeListener={unregisterChangeListener}
+                filter={this.changeFilter}
+              >
```
riccardo-forina: Same comment as above, an helper `WithUpdatingMonitoredIntegrations` component would be useful I think.

packages/api/src/Rest.tsx
gashcrumb: > Also, why are you linking the save and read functions in the state? This will mess up with the change detection done in the shouldComponentUpdate hook I think.

> If this is to pass them in the render function, just recompose the object on the fly

Gotcha!  Reworked that bit.  The state interface though has these functions defined, so not setting them yeilds a compilation error:

```
@syndesis/api: (rpt2 plugin) Error: /home/gashcrumb/Source/syndesis-react-poc/packages/api/src/Rest.tsx(39,5): semantic error TS2322 Type '{ data: T; error: false; hasData: false; loading: true; }' is not assignable to type 'Readonly<IRestState<T>>'.
@syndesis/api:   Property 'read' is missing in type '{ data: T; error: false; hasData: false; loading: true; }'.
```

I'll go with:

```
    this.read = this.read.bind(this);
    this.onSave = this.onSave.bind(this);
    this.state = {
      data: this.props.defaultValue,
      error: false,
      hasData: false,
      loading: true,
      read: this.read,
      save: this.onSave,
    };
```

Definitely needs to be done, I had head-scratching runtime errors for a bit before realizing it :-)

packages/api/src/WithServerEvents.tsx
gashcrumb: Awesome, also spotted a copypasta error I made too, phew!

syndesis/src/modules/connections/pages/ConnectionsPage.tsx
gashcrumb: Yeah, it feels like the page has to do a lot of complicated looking composition for sure.  I was meaning to update the pages that had WithPolling added to use the server event listener stuff, I'll work on adding these components instead and update these pages appropriately, thanks!

packages/api/src/WithServerEvents.tsx
gashcrumb: Yep, this is the legacy code, surprisingly removing this didn't turn up as many lint errors as I'd expect :-)

syndesis/src/modules/connections/pages/ConnectionsPage.tsx
riccardo-forina: Also please find a better name than `WithUpdatingConnections`, I suck at naming 😂

packages/api/src/Rest.tsx
gashcrumb: Ah!  Okay will do, thanks!

packages/api/src/WithChangeListener.tsx
gashcrumb: Derp, of course, thanks!  :-)

syndesis/src/modules/connections/pages/ConnectionsPage.tsx
gashcrumb: Maybe I'll rename the existing WithConnections etc to something like 'WithConnectionsInternal' and keep export the helper class instead as WithConnections?

riccardo-forina: What if we keep a single access point-eg. `WithConnections`-but add a property to enable the update strategy (that has been decided upfront by the component developer)?

So I can go and do `<WithConnection updates={true}>`, or `<WithConnection disableUpdates={true}>` if we want to have the auto-updating behavior on by default.

riccardo-forina: As a side benefit, it will enable us to disable the updating mechanism while debugging using the React Dev Tools :)

gashcrumb: Yeah, that sounds good!

gashcrumb:
@riccardo-forina I think that worked out nicely, ready for another look when you get a chance.

If you want to see it in action you can deploy an integration with the old console, then in the openshift console just scale the integration pod to 0, Syndesis will scale it back up and in the process trigger some change events.

gashcrumb:
Rebased, also added componentWillUnmount to WithServerEvents.  Still double-checking on if there's any chance the websocket could be getting leaked but can't seem to force it so will leave the browser tab open for a few days and see how it looks.

gashcrumb:
K, 1st websocket connection got closed after exactly 60 minutes, then a reconnect, so only 1 connection is held open looks like.  I'll leave this to run overnight though to be sure.

gashcrumb:
Took off the WIP label, ready to merge I think.

Assuming there isn't anything else @riccardo-forina?

riccardo-forina:
All’s good, and we can always iterate on it 👍

PR #33, opened 2018-11-28
Adds an i18n framework

elvisisking:
- uses react-i18next library
- adds syndesis/src/i18n folder for the base i18n configuration and shared translation files
- adds translation files to syndesis/src/app and syndesis/src/modules/dashboard
- App, Layout, DashboardPage, and Dashboard have been localized

Will localize rest of POC after approval of this PR. Also will add i18n instructions to developer handbook/readme.


syndesis/src/i18n/index.ts
```diff
@@ -0,0 +1,38 @@
+import i18n from 'i18next';
+import { en, it } from './locales';
+
+const options = {
+  interpolation: {
+    escapeValue: false, // not needed for react!!
+  },
+
+  debug: true, // change to false for release
```
riccardo-forina: `debug: process.env.NODE_ENV !== 'production'`

syndesis/src/i18n/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    },
+    it: {
+      app: it.it.app,
+      dashboard: it.it.modules.dashboard,
+      shared: it.it.shared,
+    },
+  },
+
+  defaultNS: 'shared',
+  fallbackLng: 'it',
```
riccardo-forina: Maybe protect this with a `process.env.NODE_ENV === 'production'` too, something like:

`fallbackLng: process.env.NODE_ENV === 'production' ? 'en' : 'it'`

riccardo-forina:
So what's your feeling about this heavy separation between the app and the packages? Does keeping all i18n concerns in a single place - the app - helps?

elvisisking:
I think the idea of keeping the "i18n concerns in a single place" is a **good** one. As you mentioned in IRC, that would definitely help if down the road we changed i18n libraries. But it would also make the maintaining of i18n-related tasks much easier.

One thing that I wanted to do differently, instead of just adding on new properties to the custom properties, was to do something like this:

``` typescript 
export interface IIntegrationsPageProps {
  linkToIntegrations: string;
  linkToIntegrationCreation: string;
  linkToConnections: string;
  ... <continue enumerating other properties here>
  i18n: { // this is the different part
    connections: string;
    linkCreateConnection: string;
    linkCreateIntegration: string;
    linkToConnections: string;
    linkToIntegrations: string;
    title: string;
    titleIntegrationUpdates: string;
  }
```
So adding an inner complex object to the properties just for i18n properties. But I ran into an issue in how to set those i18n properties in `DashboardPage` where the `Dashboard` component is constructed. Is there a way we can do that? Do you like that idea (assuming it is doable)?

I guess for me there is a somewhat bigger question ... Why have the project structured the way we do (packages and syndesis)? Is that a normal struct for React projects? For instance, having `Dashboard` in `packages` and not alongside `DashboardPage` and `DashboardApp` doesn't feel right to me. I understand a need to have a place for reusable components. Is that what is going on here?

syndesis/src/i18n/index.ts
elvisisking: Done

syndesis/src/i18n/index.ts
elvisisking: Done

syndesis/src/i18n/index.ts
```diff
@@ -22,17 +23,11 @@ const options = {
...
   defaultNS: 'shared',
-  fallbackLng: 'it',
+  fallbackLng: process.env.NODE_ENV === 'production' ? 'en' : 'it',
+  fallbackNS: ['shared'],
   keySeparator: false, // we do not use keys in form messages.welcome
-  ns: ['shared'],
+  ns: ['shared', 'app', 'dashboard'],
+} as InitOptions;
 
-  react: {
```
elvisisking: All these properties are set to default values so I just deleted them.

syndesis/src/i18n/index.ts
```diff
@@ -22,17 +23,11 @@ const options = {
   },
 
   defaultNS: 'shared',
-  fallbackLng: 'it',
+  fallbackLng: process.env.NODE_ENV === 'production' ? 'en' : 'it',
+  fallbackNS: ['shared'],
   keySeparator: false, // we do not use keys in form messages.welcome
-  ns: ['shared'],
+  ns: ['shared', 'app', 'dashboard'],
+} as InitOptions;
```
elvisisking: I had to add the `as InitOptions` statement so that `keySeparator` would work. Not sure why it worked before. Maybe it is due to upgrading the version.

syndesis/src/app/App.tsx
```diff
@@ -64,36 +65,38 @@ export class App extends React.Component {
   }
 
   public renderNavbar() {
+    const { t } = this.props;
```
riccardo-forina: Can't we use the `NamespacesConsumer` render prop here instead?

syndesis/src/i18n/index.ts
riccardo-forina: That's because before there was no proper typing for this export. This is correct.

riccardo-forina:
> @elvisisking So adding an inner complex object to the properties just for i18n properties. But I ran into an issue in how to set those i18n properties in `DashboardPage` where the `Dashboard` component is constructed. Is there a way we can do that? Do you like that idea (assuming it is doable)?

It's usually suggested to keep props flat, in order to leverage [`PureComponent`](https://reactjs.org/docs/react-api.html#reactpurecomponent)'s changes detection logic.

Namespacing props as you did - `i18nSomething` - is ok.

> @elvisisking I guess for me there is a somewhat bigger question ... Why have the project structured the way we do (packages and syndesis)? Is that a normal struct for React projects? For instance, having `Dashboard` in `packages` and not alongside `DashboardPage` and `DashboardApp` doesn't feel right to me. I understand a need to have a place for reusable components. Is that what is going on here?

One thing I noticed in these months is that we have 2 kind of issues relative to the UI: functional and UI. 

The functional ones are issues in the business logic, like some problem fetching data, or new features altogether.

The UI ones are about things the end users sees, like misplaced tooltip position, or layout changes.

With this structure if you have to work on something UI related you can just start the [Storybook](https://storybook.js.org/) for the `@syndesis/ui` package and work on the thing without having to deal with the app. So no need to have minishift and Syndesis installed, for example.
If you haven't already, try launching Storybook. From the repo root:
```bash
$ cd packages/ui
$ yarn storybook
```
A browser should open automatically pointing to [http://localhost:9001](http://localhost:9001). If not open it manually.

Functional issues will be tackled in a similar way, but they'll probably have impact on the main app and/or the api package. We don't have Storybook here, we have normal unit tests and eventually we will have [Cypress](https://cypress.io) for E2E tests.

syndesis/src/app/App.tsx
elvisisking: This method returns an array. The only way I think we can use `NamespacesConsumer` would be to wrap **each** of the 5 `PfVerticalNavItem`s and I thought that would make the code fairly ugly. Do you want me to go ahead and do that?

riccardo-forina: You should be able to change that array to use a [Fragment](https://reactjs.org/docs/fragments.html) instead (dunno why I didn't do that at the time).

```jsx
public renderNavbar() {
    return (
      <>
        <PfVerticalNavItem
          icon={'fa fa-tachometer'}
          to={'/'}
          exact={true}
          label={'Home'}
          key={1}
        />
        <PfVerticalNavItem
          icon={'pficon pficon-integration'}
          to={'/integrations'}
          label={'Integrations'}
          key={2}
        />
        <PfVerticalNavItem
          icon={'pficon pficon-plugged'}
          to={'/connections'}
          label={'Connections'}
          key={3}
        />
        <PfVerticalNavItem
          icon={'fa fa-cube'}
          to={'/customizations'}
          label={'Customizations'}
          key={4}
        />
        <PfVerticalNavItem
          icon={'pficon pficon-settings'}
          to={'/settings'}
          label={'Settings'}
          key={5}
        />
      </>
    );
```

elvisisking: Yes that works. So converted over to using NamespacesConsumer. Much better!

PR #39, opened 2018-12-04
feat: finish up the detailed integration status

### LINKED ISSUE #37: Integration detailed status UI component
* **Integration status**
  * [ ] proper implementation
  * [ ] write storybooks
  * [ ] write tests

gashcrumb:
fixes #37 - adds storybooks and tests for IntegrationProgress and IntegrationStatusDetail, and adds the log link property to both.

Also tweaks how the detailed status is fetched from the server it'll be driven by integration deployment changes vs polling.

TODO
* feedback?
* ~~rebase to pick up i18n changes~~
* ~~More manual testing with the backend to see how the detailed status works in the integration list~~
* ~~Calculate the log link somewhere and pass it in so it works~~

### DIFF
--- a/packages/ui/tests/IntegrationProgress.spec.tsx
+++ b/packages/ui/tests/IntegrationProgress.spec.tsx
@@ -1,0 +1,48 @@
+import * as React from 'react';
+import { render } from 'react-testing-library';
+import { IntegrationProgress } from '../src';
+
+export default describe('IntegrationProgress', () => {
+  // Component with a log link
+  const testComponent = (
+    <IntegrationProgress
+      value={'Deploying'}
+      currentStep={2}
+      totalSteps={4}
+      logUrl={'http://localhost:9000/'}
+      i18nLogUrlText={'View Logs'}
+    />
+  );
+
+  // Component without a log link
+  const testComponentNoLog = (
+    <IntegrationProgress
+      value={'Assembling'}
+      currentStep={1}
+      totalSteps={4}
+      i18nLogUrlText={'View Logs'}
+    />
+  );
+
+  it('Should show the progress value and steps', () => {
+    const { getByTestId } = render(testComponent);
+    expect(getByTestId('integration-progress-value')).toHaveTextContent(
+      'Deploying ( 2 / 4 )'
+    );
+  });
+  it('Should show the log link when supplied', () => {
+    const { queryByTestId } = render(testComponent);
+    expect(queryByTestId('deployment-log-link')).toBeDefined();
+    expect(queryByTestId('deployment-log-link')).toHaveTextContent('View Logs');
+  });
+  it('Should show the progress value and steps', () => {
+    const { getByTestId } = render(testComponentNoLog);
+    expect(getByTestId('integration-progress-value')).toHaveTextContent(
+      'Assembling ( 1 / 4 )'
+    );
+  });
+  it('Should not show the log link if not supplied', () => {
+    const { queryByTestId } = render(testComponentNoLog);
+    expect(queryByTestId('deployment-log-link')).toEqual(null);
+  });
+});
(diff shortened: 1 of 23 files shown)


packages/ui/tests/IntegrationProgress.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+  });
+  it('Should show the progress value and steps', () => {
+    const { getByTestId } = render(testComponentNoLog);
+    expect(getByTestId('integration-progress-value')).toHaveTextContent(
+      'Assembling ( 1 / 4 )'
+    );
+  });
+  it('Should not show the log link if not supplied', () => {
+    const { getByTestId } = render(testComponentNoLog);
+    const logSpan = getLogSpan('deployment-log-link', getByTestId);
```
riccardo-forina: Every `getXXX` has a [`queryXXX` counterpart](https://github.com/kentcdodds/react-testing-library#query-apis) that can be used to verify for the absence of an element.

You should be able to simplify this using a `queryByTestId` and then check for `null`

gashcrumb: ah, excellent, thanks!

gashcrumb:
Okay, ready for review, *except* since I can't get a minishift running I haven't been able to test and verify that it behaves the way it should.  But it should :-)

syndesis/src/i18n/locales/shared-translations.en.json
```diff
@@ -17,7 +17,10 @@
       "linkCreateConnection": "Create Connection",
       "linkCreateIntegration": "Create Integration",
       "Name": "Name",
+      "Pending": "Pending",
       "Published": "Published",
+      "podLogsUrl": "{{consoleUrl}}/project/{{namspace}}/browse/pods/{{podId}}?tab=logs",
+      "podEventsUrl": "{{consoleUrl}}/project/{{namespace}}/browse/pods/{{podId}}?tab=events",
```
gashcrumb: @elvisisking Wonder, am I abusing the localization here?  We followed this pattern in the old app, was nice to have this externalized a bit.

elvisisking: If I understand the question, then yes, I think you might be misusing localization here. You are talking about the `podLogsUrl` and `podEventsUrl` keys right? These localization keys don't need any translations do they? Are you just using them for data substitution? Those 2 keys don't appear in the Italian language translations file either. And also the `viewLogs` key seems to be missing here?

gashcrumb: Damn, forgot `viewLogs`, thanks!

Yeah, I'll rework how these URLs are built up instead.  Though generally we may want to think about how to externalize these and also other things like the various documentation URLs that we'll eventually have to have.

gashcrumb: created https://github.com/syndesisio/syndesis-react-poc/issues/44 for that and sorted the above, thanks!

gashcrumb:
It works!  Still going to need some styling refinement but it's now working the same as the legacy one:

image

gashcrumb:
@riccardo-forina also found if I just run `yarn minishift:restore` *after* starting up `yarn watch:app` I can still get at the old app and do stuff yet also see the react app via `http://localhost:3000` and it seems to work as normal.

gashcrumb:
Styling sorted, ready to merge.

riccardo-forina:
One improvement I’d suggest is to explicitly define the values the current and target state can have in the component’s prop ( something like ‘this’ | ‘that’ | ‘something’) and perhaps expose them as constants so that we have fewer magic strings around.

riccardo-forina:
Mind, this wouldn’t save us from new unknown states coming from the BE since if I’m not mistaken the type is simply string?, but it would tell the future us that the component was built with just those states in mind

gashcrumb:
Done!  I wound up having to add a `sharedModels.ts` file since multiple components refer to the integration state field there.  I wonder though if eventually we should try and move some of the logic for pending/published/unpublished out of the UI component and into the app.  Maybe not in this PR though :-)

PR #42, opened 2018-12-05
chore: move the app layout to @syndesis/ui, and test it

### LINKED ISSUE #38: Move the app layout to @syndesis/ui

riccardo-forina:
Closes #38


packages/ui/.storybook/addons.js
```diff
@@ -1,3 +1,4 @@
 import '@storybook/addon-a11y/register';
 import '@storybook/addon-knobs/register';
 import '@storybook/addon-options/register';
+import '@storybook/addon-actions/register';
```
riccardo-forina: Added but not used. I still kept it since it will be handy in the future.

packages/ui/.storybook/config.js
```diff
@@ -47,7 +47,7 @@ addDecorator(
      * show addon panel as a vertical panel on the right
      * @type {Boolean}
      */
-    addonPanelInRight: true,
+    addonPanelInRight: false,
```
riccardo-forina: This moves the tools panel to the bottom, which is a bit handier if your browser isn't running fullscreen. This can be reverted if we liked the tool on the right more.

syndesis/src/app/App.tsx
```diff
@@ -29,50 +40,76 @@ export interface IAppBaseProps {
...
+                        headers: { 'SYNDESIS-XSRF-TOKEN': 'awesome' },
+                      }}
+                    >
+                      <ApiContext.Consumer>
+                        {({ apiUri, headers }) => (
+                          <WithServerEvents apiUri={apiUri} headers={headers}>
+                            {functions => (
+                              <ServerEventsContext.Provider value={functions}>
+                                <AppLayout
+                                  appTitle={'Syndesis'}
```
riccardo-forina: @gashcrumb @elvisisking we handle the app name in a special way in the Angular app. What should we do here?

gashcrumb: In the angular app it's in the localization file, but then we override it by setting it from the `config.json`, which then lets us easily override the name for the productized version.  We should do something similar here I think, or at least pull it from the `config.json` for starters.

Might be worth doing a stopgap solution for now and we can create an issue and collect requirements maybe?

gashcrumb: heh, kicked the can down the road :-)

Issue #3, opened 2018-10-31
Dynamic forms


gashcrumb:
In a few places in the Syndesis app we leverage the ng-dynamic-forms component to produce a set of fields for the user to fill in from some JSON that's part of the connector/connection definition.

So it'd be good to investigate/POC what approaches we could do in this new codebase.

gashcrumb:
I dunno, it seems like it might be easiest to roll our own for this vs using a third party package, thoughts?

gashcrumb:
So we're thinking that JSON -> forms won't be a thing.

gashcrumb:
Thinking about this more, there's a chance we could need some subset of this kind of functionality still but maybe in more limited cases.  The specific case is the Salesforce connector, several of it's actions require more than one "page" of configuration, i.e. data from the first form presented to the user influences the next form that the user sees.

gashcrumb:
Also I think we want to avoid bringing in a 3rd party library to do forms and instead be something we do specific to syndesis, as we'll have *much* less concerns than a general purpose form library would.

gashcrumb:
So we have this now, will have to create issues to address gaps.

PR #43, opened 2018-12-05
Dynamic form generation

seanforyou23:
This PR adds some very basic support for dynamic forms. There is a form builder function, which, when provided with JSON structures similar to those found already in Syndesis - for example;

`{
  clientSecret: {
    description: '',
    displayName: '',
    type: 'string',
    labelHint: ''
  },
  authorizationEndpoint: {
    description: '',
    displayName: '',
    type: 'string',
    labelHint: ''
  }
}`

produce a static form, for example like the below screenshot.

screen shot 2018-12-05 at 3 07 24 pm

The idea is that these form utility functions will work together similar to how Angular's FormBuilder class works, just less all the angular specific stuff. So when we move from PF3 to PF4, we might be able to leave much of the form builder still intact.

There is much work left to do, so I won't bother enumerating all of it here. Instead, for now, feel free to check out this work and provide any general feedback you may have in terms of high-level design patterns, directory structure, or naming contentions to enforce.


packages/ui/src/FormBuilder/FormBuilder.tsx
```diff
@@ -0,0 +1,52 @@
...
+  );
+};
+
+// formGroup renders a group of related form elements
+const formGroup = ({ formControlProperty, children }: any): JSX.Element => {
+  return <Form.FormGroup key={formControlProperty}>{children}</Form.FormGroup>;
+};
+
+// formBuilder takes a properties object and dynamically builds a patternfly react form
+export const formBuilder = (properties: object): JSX.Element => {
```
gashcrumb: Maybe make an interface for properties, even if it's like 

```typescript
interface IFormProperties { 
  [name: string]: ISynFormControl; 
}
```

packages/ui/src/FormBuilder/FormUtils.tsx
```diff
@@ -0,0 +1,133 @@
+import { Form } from 'patternfly-react';
+import * as React from 'react';
+
+export interface ISynFormControl {
```
gashcrumb: Wonder this probably should go in the models package eventually?  Along with whatever we call the interface for the `properties` object that we pass in.

packages/ui/.storybook/StoryHelper.css
```diff
@@ -13,7 +13,5 @@ body,
 }
 
 .test-area {
-  align-items: center;
```
gashcrumb: Does this mess up the other storybook views?  Probably better to avoid this and add some css specific to the form builder component to make it behave appropriately.

gashcrumb:
If we like the direction I'd vote for merging this work sooner rather than later just so we can have basic functionality to start fleshing out dependent bits like the create connection flow, editor etc.

@riccardo-forina WDYT?

packages/ui/src/FormBuilder/FormUtils.tsx
gashcrumb: Ah, you know what, there's actually a `ConfigurationProperty` interface in the `models` package, we should use that one I think.

seanforyou23: Good point. So that would be the only place where the ui package is dependent on the models package from what I can see. @riccardo-forina that cool? When I try to add `@syndesis/models` using yarn I get errors (probably just doing it wrong). I guess I could even see the UI package getting its own set of models, but I'm curious what others think.

packages/ui/.storybook/StoryHelper.css
seanforyou23: Totally does, I was planning to put that back.. now seems like a good time :)

packages/ui/src/FormBuilder/FormUtils.tsx
gashcrumb: Ah of course.  We probably don't want that dependency now that you mention it.

riccardo-forina: Yes, better not to add a dependency to the models in the UI package, I kind of think of it like a spy that tells me that I'm doing something wrong.

riccardo-forina:
Thank you @seanforyou23, nice progress!

I think this is an interesting approach, that now requires thinking about wiring the components for state management.

In the past I mentioned [Formik](https://jaredpalmer.com/formik) as a potential candidate for form handling, and looking at this PR and how Formik works I still think it would be a nice fit.

I think we could try to achieve something like this:

```jsx
<FormBuilder properties={properties}>
  ({ initialValues, validationSchema, getFields })
  <Formik
    initialValues={initialValues}
    validationSchema={validationSchema}
    onSubmit={values => {
      // same shape as initial values
      console.log(values);
    }}
  >
    {({ errors, touched, values }) => (
      <Form>
        {/* 
          getFields will return the list of fields that should be rendered, based on the 
          current values (so to resolve the field relationships), ordered as defined by 
          the order property.  
        */}
        {getFields(values).map((field) => (
          <Field 
            name={field.name} 
            component={field.component} 
            key={field.name} 
          />
          {errors[field.name] && touched[field.name] ? (
            <div>{errors[field.name]}</div>
          ) : null}
        ))}
        <button type="submit">Submit</button>
      </Form>
    )}
  </Formik>
</FormBuilder>
```
_Disclaimer: totally untested code, it might bring destruction in the multiverse if run_

So, different from this PR there is the fact that `formBuilder` is a render prop component to allow composition, and `field.component` will refer to custom-made components that will use PF for rendering and will be Formik-aware (but still decoupled from it). Example component as by the doc:

```jsx
const CustomInputComponent = ({
  field, // { name, value, onChange, onBlur }
  form: { touched, errors }, // also values, setXXXX, handleXXXX, dirty, isValid, status, etc.
  ...props
}) => (
  <div>
    <input type="text" {...field} {...props} />
    {touched[field.name] &&
      errors[field.name] && <div className="error">{errors[field.name]}</div>}
  </div>
);
```

WDYT?

seanforyou23:
Thanks for the feedback, definitely helps. I'm looking into the Formik stuff, and also just getting familiar with the render prop pattern, that seems to be key in all of this. I'll send more comments as I get my head wrapped around all of this. Generally, sounds like a good direction though!

gashcrumb:
Related to #3

Issue #41, opened 2018-12-05
Customization: API Client Connectors page layout


elvisisking:
1. list

- implementation of components needed for the list view
- write storybooks
- write tests

2. list empty page layout

- implementation
- write storybooks
- write tests

PR #55, opened 2018-12-20
More work on building out customizations

elvisisking:
- Work related to issue #41 
- replace placeholder implentation of CustomizationsExtensionDetail with real one
- updated CustomizationsExtensionDetail story
- deleted duplicate dependency in syndesis/package.json
- added utility functions to work with OptionalInt type
- changed 'stacked' property to 'true' in CustomizationsApiConnectorListItem and Customization/CustomizationsExtensionListItem for better readability


elvisisking:
@riccardo-forina I don't know how to fix the build error I'm getting on [this line](https://github.com/elvisisking/syndesis-react-poc/blob/extension-detail-screen/packages/ui/src/Customization/CustomizationsExtensionDetail.tsx#L178). The error says that `rowData` has an implicit type of `any`. But when I try and add a type like this:
```
(value: string, { rowData: Extension }) => (
```
I get this error in the editor: `'Extension' is declared but its value is never read.`

Any ideas?  Thanks!

elvisisking:
Changing the offending line above to this:
```
(value: string, { rowData }: { rowData: any }) => (
```
got the build to work @riccardo-forina.

packages/ui/package.json
```diff
@@ -67,6 +67,8 @@
     "react-router-dom": "^4.3.1"
   },
   "dependencies": {
+    "@syndesis/models": "*",
```
gashcrumb: I think we wanted to avoid binding the UI widgets to the model ideally.  I think this was added to support IExtensionDetailProps `extension` member but I kinda recall we want to let the UI widgets have their own interface/props that's decoupled from the backend model.

packages/ui/src/Customization/CustomizationsExtensionDetail.tsx
```diff
@@ -1,14 +1,249 @@
...
+  Table,
+  Tooltip,
+} from 'patternfly-react';
 import * as React from 'react';
+import './CustomizationsExtensionDetail.css';
 
 export interface IExtensionDetailProps {
-  extensionId: string;
-  i18nTitle: string;
+  extension: Extension;
```
gashcrumb: [this comment](https://github.com/syndesisio/syndesis-react-poc/pull/55/files#r245312397) I *think* we want to avoid referring to the model-generated type and instead create a separate interface or ideally directly require what properties of an extension this widget supports, vs passing in the whole extension.

packages/ui/src/Customization/CustomizationsExtensionDetail.tsx
```diff
@@ -1,14 +1,249 @@
...
+
+    if (this.props.extension.extensionType === 'Steps') {
+      supportedMsg = this.props.i18nSupportedSteps;
+    } else if (this.props.extension.extensionType === 'Connectors') {
+      supportedMsg = this.props.i18nSupportedConnectors;
+    }
+
+    return (
+      <div className="extension-group">
+        <h3>{supportedMsg}</h3>
```
gashcrumb: What's shown here for a "library" extension?

packages/ui/src/Customization/CustomizationsExtensionDetail.tsx
```diff
@@ -1,14 +1,249 @@
...
+          <dd>{this.props.i18nLastUpdateDate}</dd>
+        </dl>
+      </div>
+    );
+  }
+
+  public createSupportsSection() {
+    let supportedMsg;
+
+    if (this.props.extension.extensionType === 'Steps') {
```
gashcrumb: Should probably create constants for these strings or an enum type for this extensionType field maybe?

packages/ui/src/Customization/CustomizationsExtensionDetail.tsx
```diff
@@ -1,14 +1,249 @@
...
+  i18nType: string;
+  i18nTypeMessage: string;
+  i18nUpdate: string;
+  i18nUpdateTip?: string;
+  i18nUsage: string;
+  i18nUsageMessage: string;
+  onDelete: (extensionName: string) => void;
+  onSelectIntegration: (integrationId: string) => void;
+  onUpdate: (extensionName: string) => void;
+  propDescription: string;
```
gashcrumb: What's this field for?  I feel like this could use a better name or comment.

packages/ui/src/Customization/CustomizationsExtensionDetail.tsx
```diff
@@ -1,14 +1,249 @@
...
+  propName: string;
```
gashcrumb: Same with this field.

packages/ui/src/Customization/CustomizationsExtensionDetail.tsx
gashcrumb: [Here's](https://github.com/syndesisio/syndesis-react-poc/blob/master/packages/ui/src/Integration/IntegrationsListItem.tsx#L8-L9) a little example, instead of passing in the integration object, the consumer of this component passes in the ID, name, state etc as separate fields.  This helps ensure the UI bits are decoupled from the generated model, and the mapping of the type in the response is handled by the parent component.  

So then for this PR, the component wouldn't accept an "Extension" but all the fields that it needs to display, ideally.

packages/ui/src/Customization/CustomizationsExtensionDetail.tsx
gashcrumb: Maybe create a `models.ts` file for this kind of stuff, this was done for example in the [integrations](https://github.com/syndesisio/syndesis-react-poc/tree/master/packages/ui/src/Integration) UI module.

packages/ui/src/Customization/CustomizationsExtensionDetail.tsx
elvisisking: Good catch. In the angular version the `h3` is not added for library extension. I will change so that the `h3` is not added for this case.

packages/ui/src/Customization/CustomizationsExtensionDetail.tsx
elvisisking: Agree. I'll add a `models.ts` file.

packages/ui/src/Customization/CustomizationsExtensionDetail.tsx
elvisisking: Agree. I'll come up with better names for this and `propName` (as soon as I figure out what the heck those properties are:)

packages/ui/src/Customization/models.ts
```diff
@@ -0,0 +1,27 @@
+// Extension Types
+export enum ExtensionType {
+  Connectors = 'Connectors',
+  Libraries = 'Libraries',
+  Steps = 'Steps',
+}
+
+export interface IAction {
```
gashcrumb: This interface and `IIntegration` feel like types that we don't need to define, at least in this way, *however* I don't yet have a better suggestion at the moment :-)

packages/ui/src/Customization/CustomizationsExtensionDetail.css
```diff
@@ -0,0 +1,16 @@
...
+.extension-id {
+  font-size: 0.5em;
+  margin-left: 10px;
+}
+
+.extension-title {
+  margin: 0;
+}
+
+button + button {
```
riccardo-forina: Can you refactor these classes to use BEM, and namespace that button + button thing?

elvisisking:
I have moved the work on the extension detail UI to a new branch so this PR can be closed. I have addressed the issues brought up by @riccardo-forina and @gashcrumb in the new branch.

PR #56, opened 2019-01-17
Initial Implementation of Customization API Client Connectors and Extensions Page

elvisisking:
- more work related to #41 
- added WithApiConnectors and WithExtensions
- added ApiConnectorsPage and ExtensionsPage
- added OptionalIntUtil to work with OptionalInt type
- deleted CustomizationsPage as it is no longer needed
- fixed some bugs in CustomizationsApiConnectorListItem and CustomizationsExtensionListItem

**Note: Still to implement: import extension, create API connector, detail pages, delete handlers, and update extension handler**


packages/api/src/WithApiConnectors.tsx
```diff
@@ -0,0 +1,57 @@
...
+export interface IWithApiConnectorsProps {
+  disableUpdates?: boolean;
+  children(props: IRestState<IApiConnectorsResponse>): any;
+}
+
+export class WithApiConnectors extends React.Component<
+  IWithApiConnectorsProps
+> {
+  public changeFilter(change: IChangeEvent) {
+    return change.kind.startsWith('apiConnector');
```
riccardo-forina: @gashcrumb do we have a list of values that `kind` can take somewhere?

packages/ui/src/Customization/CustomizationsExtensionListItem.tsx
```diff
@@ -28,6 +29,54 @@ export interface IExtensionListItemProps {
...
+    return (
+      <Tooltip id="updateTip">
+        {this.props.i18nUpdateTip
+          ? this.props.i18nUpdateTip
+          : this.props.i18nUpdate}
+      </Tooltip>
+    );
+  }
+
+  public handleDelete = () => {
```
riccardo-forina: So, I prefer not using fat arrows to define class methods, because you don't get class methods this way, meaning that the method will not be present in the class prototype.

Is this a big deal? It depends, [this article](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1) explains it in details.

As a rule of thumb, since this approach can lead to performance issues with many instances of the same component, I always write "normal" class methods and do the bind in the constructor manually. But that's me.

packages/utils/src/OptionalIntUtil.ts
```diff
@@ -0,0 +1,37 @@
+import { OptionalInt } from '@syndesis/models';
+
+export class OptionalIntUtil {
```
riccardo-forina: I'd dump the class since it contains only static methods, and export the functions directly.

syndesis/src/modules/customizations/customizations.css
```diff
@@ -0,0 +1,21 @@
+.bottom-bordered {
```
riccardo-forina: It would be better not to have CSS in the main app, unless we are talking about app-wide utility classes.

These look like are specific to the customization section, so the components using them should be refactored to be exported by the syndesis/ui package, and classes renamed to leverage the BEM methodology. The convention I have been following is `.{component-name}__{component-item}` (eg. `.content-with-sidebar__sidebar`).

packages/api/src/WithApiConnectors.tsx
elvisisking: Actually I just *guessed* at the value of `change.kind` just to put something in there. Testing things confirms that my guess was not a good one. I looked at the angular codebase and couldn't find the answer (maybe you can @gashcrumb). On the upside, testing the change event in `WithExtensions` works fine.

packages/ui/src/Customization/CustomizationsExtensionListItem.tsx
elvisisking: I'll make that change. Also making same change in `CustomizationsApiConnectorListItem`. Also in `CustomizationsApiConnectorListItem` will make `render` that last method in the file.

packages/utils/src/OptionalIntUtil.ts
elvisisking: I'll remove the class and export the 3 functions.

syndesis/src/modules/customizations/customizations.css
elvisisking: I created `NavLinkTab` in ui/Shared to address this comment. Wdyt?

elvisisking:
I have a second commit that addresses your comments @riccardo-forina. Take a look. Thanks for your comments.

packages/api/src/WithApiConnectors.tsx
gashcrumb: @riccardo-forina  I'll have to make one most likely.  Effectively it's not really bound to a set list.

@elvisisking I think it'd be `Connectors` as once an API client connector is created, it's just a connector.  But it has some special metadata, *however* I don't think this metadata is available within the IChangeEvent object you get back, typically they just have the kind and an ID of the changed thing.

syndesis/src/modules/customizations/customizations.css
elvisisking: @riccardo-forina I forgot to delete the `customizations.css file`. I'll amend the last commit and let you know.

gashcrumb:
@elvisisking does this PR need #55 merged first?

packages/ui/src/Customization/CustomizationsExtensionListItem.tsx
```diff
@@ -59,23 +59,23 @@ export class CustomizationsExtensionListItem extends React.Component<
     );
   }
 
-  public handleDelete = () => {
+  public handleDelete() {
```
riccardo-forina: `this` here will not be what you'd expect to be since this method hasn't bind bound to the class instance in the constructor.

packages/ui/src/Shared/NavLinkTab.css
```diff
@@ -0,0 +1,16 @@
+.NavLinkTab {
+  color: black;
+  font-size: 14px;
+}
+
+.NavLinkTab_disabled {
```
riccardo-forina: Modifiers should be expressed like this: `.NavLinkTab.is-disabled`

packages/ui/src/Shared/NavLinkTab.tsx
```diff
@@ -0,0 +1,38 @@
...
+    if (this.props && this.props.disableLink) {
+      event.preventDefault();
+    }
+  }
+
+  public render() {
+    return (
+      <NavLink
+        className={
+          this.props.disableLink ? 'NavLinkTab_disabled' : 'NavLinkTab'
```
riccardo-forina:
```tsx
import classnames from 'classnames';

<NavLink className={classnames('NavLinkTab', { 'is-disabled': this.props.disableLink })} ...
```

packages/ui/src/Shared/NavLinkTab.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  // The URL string to use when the link is clicked.
+  toLink: string;
+}
+
+// A class to represent a tab. Use multiple to create a tab folder. A tab can be disabled when its
+// link should not be navigated too.
+export class NavLinkTab extends React.Component<INavLinkTabProps> {
+  public handleClick(event: React.MouseEvent<HTMLAnchorElement>) {
+    if (this.props && this.props.disableLink) {
```
riccardo-forina: This method isn't bound, `this` will not point to the class instance.

packages/api/src/WithApiConnectors.tsx
elvisisking: Looks like "connector" works @gashcrumb. When an API connector is created or deleted the ChangeEvent.kind is that. I'll make that change.

elvisisking:
No @gashcrumb. #55 is not needed for this one.

packages/ui/src/Shared/NavLinkTab.css
elvisisking: Changed name in css file as you recommend.

packages/ui/src/Shared/NavLinkTab.tsx
elvisisking: I updated component with your change.

packages/ui/src/Shared/NavLinkTab.tsx
elvisisking: Added constructor and method bind statement.

elvisisking:
Check latest commit out @riccardo-forina. I believe I addressed all your latest comments.

riccardo-forina:
@elvisisking nice! What about the changes [here](https://github.com/syndesisio/syndesis-react-poc/pull/56/files/8226b2d5309c60b23123c22610ae90c81afee9dc..8e71a0992a696787e7358cdd11e5db41ea623dee#diff-4773aff2d2ad90e351c8497fa8ad9937)?

PR #57, opened 2019-01-25
Integration editor

riccardo-forina:
*Warning: massive PR*

This implements the integration editor, for both the creation and editing flows for normal integration (read, not for API Provider ones).

It contains a number of bug-fixes, refactoring and rethinking of many components across all packages.
It's also totally untested but works well enough for the moment.

Work isn't finished yet, but I'm not sure it's a good idea to add even more changes to the codebase while syncing with you guys.

WDYT?


packages/api/src/WithIntegrationHelpers.tsx
```diff
@@ -0,0 +1,352 @@
...
+  }
+
+  public makeLocalStorageId(id: string) {
+    return `iec-${id}`;
+  }
+
+  public async createDraft(value: Integration) {
+    // TODO: this should be handled by the BE
+    if (!value.id) {
+      throw new Error("Integration has no id, can't create a draft");
```
gashcrumb: We *could* create this ID easily enough [this here](https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/platform/helpers/key.ts) is how it's done in the angular codebase.

packages/api/src/integrationHelpers/helpers.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export function getActionsWithTo(actions: Action[] = []) {
+  return actions.filter(a => a.pattern === 'To');
+}
+
+export function getConnectionConnector(
+  connection: ConnectionOverview
+): Connector {
+  if (!connection.connector) {
+    throw Error(`FATAL: Connection ${connection.id} doesn't have a connector`);
```
gashcrumb: Good call, we should probably do more sanity checks like this.

packages/ui/src/Integration/IntegrationFlowAddStep.tsx
```diff
@@ -0,0 +1,101 @@
...
+        </div>
+        {this.props.showDetails && this.props.children}
+        <Overlay
+          placement="top"
+          rootClose={false}
+          show={this.props.forceTooltip || this.state.showTooltip}
+          target={this.containerRef.current}
+          container={this}
+        >
+          <Popover
```
gashcrumb: We should consider combining steps and connections in this codebase and not bring forward the separate pages/options like we have in the angular codebase.  It's something we've wanted to do anyways, may as well simplify this and consider doing it here.  Then this whole popover can go away for example :-)

packages/api/src/WithIntegrationHelpers.tsx
gashcrumb: Rofl, and here I find the file in this PR :-)

riccardo-forina: Actually, that file and the draft methods are leftovers that I forgot to remove! 

So the idea to keep the integration state - before me realizing that the history could do that for me - was to call an API to store the draft and retrieve it when needed. These helpers were supposed to do that and since we have no such API the plan was to fall back on the localStorage. 

I'll do some cleanup and remove the unused stuff.

packages/auto-form/src/widgets/FormInputComponent.tsx
```diff
@@ -3,21 +3,26 @@ import * as React from 'react';
...
-  form: { touched, errors },
+  form: { touched, errors, isSubmitting },
   ...props
 }: {
   [name: string]: any;
 }) => (
   // TODO replace with PF3/PF4 widget
   <div className="form-group">
-    <label className="col-sm-2 control-label" htmlFor={field.name}>
+    <label className="ontrol-label" htmlFor={field.name}>
```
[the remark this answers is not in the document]
riccardo-forina: Yup!

packages/ui/src/Integration/IntegrationFlowAddStep.tsx
riccardo-forina: So merge steps and connections together? Actually, this is something I had to ask, what are steps and how do they work? Are hard-coded somehow?

packages/api/src/SyndesisFetch.tsx
```diff
@@ -1,27 +1,28 @@
 import * as React from 'react';
 import { ApiContext } from './ApiContext';
-import { IRestRenderProps, Rest } from './Rest';
+import { IFetchRenderProps, Fetch } from './Fetch';
 import { Stream } from './Stream';
 
-export interface ISyndesisRestProps<T> {
+export interface ISyndesisFetchProps<T> {
```
elvisisking: Does it make sense for this to extend `IFetchProps`?

packages/api/src/Fetch.tsx
```diff
@@ -3,65 +3,54 @@ import * as React from 'react';
...
 }
 
-export interface IRestProps<T> {
-  autoload?: boolean;
+export interface IFetchProps<T> {
   baseUrl: string;
   url: string;
   headers?: IFetchHeaders;
   contentType?: string;
   defaultValue: T;
```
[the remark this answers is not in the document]
riccardo-forina: I admit the reason behind this is kinda personal, I don't like having to write the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) across the codebase, or be forced to handle the loading state.

So for example [here](https://github.com/riccardo-forina/syndesis-react-poc/blob/editor-alternative-ui/syndesis/src/modules/dashboard/pages/DashboardPage.tsx#L150) without a default value I should have wrapped the `ConnectionsMetric` component with a `WithLoader` helper as I did [here](https://github.com/riccardo-forina/syndesis-react-poc/blob/editor-alternative-ui/syndesis/src/modules/dashboard/pages/DashboardPage.tsx#L218), but it feels overkill.

But I'm open to change it to be optional if we prefer.

packages/api/src/SyndesisFetch.tsx
riccardo-forina: Since `SyndesisFetch` is about simplifying the usage of the underlying `Fetch` component I'd say no.

packages/api/src/WithConnections.tsx
```diff
@@ -1,19 +1,58 @@
...
+
+export function getConnectionsWithFromAction(
+  connections: ConnectionOverview[]
+) {
+  return connections.filter(connection => {
+    if (!connection.connector) {
+      // safety net
+      return true;
+    }
+    return connection.connector.actions.some(action => {
```
elvisisking: Could you use `getActionsWithFrom` in `helpers.tsx` here?

packages/api/src/WithConnections.tsx
```diff
@@ -1,19 +1,58 @@
...
+  return connections.filter(connection => {
+    if (!connection.connector) {
+      // safety net
+      return true;
+    }
+    if (connection.connectorId === 'api-provider') {
+      // api provider can be used only for From actions
+      return false;
+    }
+    return connection.connector.actions.some(action => {
```
elvisisking: Could you use `getActionsWithTo` in `helpers.tsx` here?

packages/api/src/WithIntegrationHelpers.tsx
```diff
@@ -0,0 +1,352 @@
...
+      saveIntegration: this.saveIntegration,
+      setDraft: this.setDraft,
+      setName: this.setName,
+      updateConnection: this.updateConnection,
+      updateOrAddConnection: this.updateOrAddConnection,
+    });
+  }
+}
+
+export const WithIntegrationHelpers: React.FunctionComponent<
```
elvisisking: Just a general question (I think we actually talked about this), where/how are all the errors thrown within this component handled?

riccardo-forina:
I went through most of the code and added comments to explain the component/function intent and mark as many TODO actions as I could remember/think of.

@elvisisking thank you for your feedback, I'll go through it tomorrow!

PR #63, opened 2019-02-11
Fix Storybook build errors

kahboom:
To resolve all but one of the Storybook build errors. See comment and screenshot below.


kahboom:
FYI these are the errors:

screenshot 2019-02-11 09 26 27

Only one missing is for `AppLayout.stories.tsx` since the properties `onNavigationCollapse` and `onNavigationExpand` are not listed in the story, but are declared as properties in the interface. We need to use storybook addon actions to reference them from the knob thingy.

packages/ui/stories/Layout/AppLayout.stories.tsx
```diff
@@ -1,7 +1,7 @@
+import { text } from '@storybook/addon-knobs';
```
kahboom: This was to get rid of the warning about imports not being in order.

PR #64, opened 2019-02-12
fix(storybook): unbreak storybook for ui pkg

seanforyou23:
This PR unbreaks the UI packages storybook. There may better ways of fixing the issues, but injecting MemoryRouter and adding ` || ''` for href value in connection card story at least allows the UI to fully render.


kahboom:
Thanks for looking into this @seanforyou23 ! Think @riccardo-forina would be better-suited to give the final OK on it, I'm interested in seeing what he thinks.

riccardo-forina:
Thank you!

PR #73, opened 2019-02-20
TEIIDTOOLS-567

mdrillin:
Adds Data tab to syndesis-react-poc and adds the virtualization summary page
- DataApp is added similar to other modules
- VirtualizationsPage is added, which displays the available virtualizations.  This page utilizes the components previously added by blafond.
- The associated i18 files were added.  Some commonly used terms were added to the shared i18, while others specific to data were added to the data i18n files.  (Entries were made in the italian translations, but the actual translation values will be need to be updated).
- dvUrl was added to the config.json, since the rest interfaces for DV (teiid-komodo) are different. App.tsx was modified to include the dvApiUri
- WithVirtualizations was added to provide the virtualizations for the summary page.  Verified success in retrieving from the deployed komodo service.


packages/api/src/callDVFetch.ts
```diff
@@ -0,0 +1,34 @@
...
+      Accept: accept,
+      'Content-Type': contentType,
+      ...headers,
+    },
+    method,
+    mode: 'cors',
+    redirect: 'follow',
+    referrerPolicy: 'no-referrer',
+  });
+}
```
riccardo-forina: Is this function needed? The original `callFetch` function is a low-level one that doesn't know anything about the app, it can be safely used for DV concerns too.

packages/ui/src/Data/Virtualizations/VirtListView.tsx
```diff
@@ -61,16 +61,18 @@ export class VirtListView extends React.Component<IVirtsListViewProps> {
...
+        </div>
         <ListViewToolbar {...this.props}>
           <div className="form-group">
             <OverlayTrigger
               overlay={this.getImportVirtTooltip()}
               placement="top"
             >
               <Button
-                psStyle="default"
+                psstyle="default"
```
riccardo-forina: This should be `bsStyle` I think: https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-3/patternfly-react/src/components/Button/Button.stories.js#L25

PR #74, opened 2019-02-21
feat: upgrade the connections module to use the routes/resolvers pattern.

riccardo-forina:
These are the changes done during the live share session, plus a little bit of refactoring and cleanup.

### DIFF
--- a/syndesis/src/modules/connections/pages/ConnectorFormPage.tsx
+++ b/syndesis/src/modules/connections/pages/ConnectorFormPage.tsx
@@ -9,7 +9,7 @@ export interface IConnectorFormPageRouteParams {
 }
 
 export interface IConnectorFormPageRouteState {
-  connector: Connector;
+  connector?: Connector;
 }
 
 export default class ConnectorFormPage extends React.Component {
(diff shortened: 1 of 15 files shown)


syndesis/src/modules/connections/pages/ConnectorFormPage.tsx
```diff
@@ -1,16 +1,26 @@
...
 import { Spinner } from 'patternfly-react';
 import * as React from 'react';
 import { WithConnectorCreationForm } from '../containers';
 
+export interface IConnectorFormPageRouteParams {
+  connectorId: string;
+}
+
+export interface IConnectorFormPageRouteState {
+  connector: Connector;
```
elvisisking: Can this be optional?

syndesis/src/modules/dashboard/pages/DashboardPage.tsx
```diff
@@ -132,7 +132,7 @@ export default () => (
                       linkToIntegrations={resolvers.integrations.list()}
                       linkToIntegrationCreation={resolvers.integrations.create.start.selectConnection()}
                       linkToConnections={'/connections'}
```
elvisisking: Should you use `resolvers` for this too?

syndesis/src/modules/connections/pages/ConnectorFormPage.tsx
riccardo-forina: I guess it could, yep!

syndesis/src/modules/dashboard/pages/DashboardPage.tsx
riccardo-forina: I should indeed 😂

PR #77, opened 2019-02-22
chore: remove direct dependency on patternfly from the syndesis app

riccardo-forina:
## Changes

* Add a Tslint rule to check for unused React components interface props. This doesn't work with FunctionalComponents for the moment. A PR to fix this problem [has been submitted](https://github.com/Microsoft/tslint-microsoft-contrib/pull/824), we have to wait for a new release of the tool.
* Drop direct styling of buttons in favour of the ButtonLink component. We had many `className={'btn btn-something'}` around the code, and mixed usages of `a`, `button` and `Link` components. Now we can handle all scenarios with `ButtonLink`.
* No more `<div className='container-fluid'>...</div>`, welcome `<Container>...</Container`.
* Remove direct dependency on `patternfly-react` from the app.


riccardo-forina:
This should be ok to be reviewed. I updated the description to give you an idea of what happened.

kahboom:
I think this looks good, but I'm afraid to approve it. 👀

packages/ui/src/Data/Virtualizations/VirtListSkeleton.tsx
```diff
@@ -1,3 +1,5 @@
+// tslint:disable react-unused-props-and-state
```
kahboom: Can we not put this in the global settings?

riccardo-forina: You mean [this line](
https://github.com/syndesisio/syndesis-react-poc/pull/77/files#diff-ace19bd0c04529e685320269e3c05de9R18)? That would remove the rule altogether, which is kinda meh... I found a few problematic areas thanks to it

kahboom: I figured we could add it to the tslint config file, but nevermind. 👍

Issue #78, opened 2019-02-25
Customization: Extensions page layout


elvisisking:
Implementation of the UI related to extensions.

### list
* [x]  implementation
* [x]  write storybooks
* [x]  write tests

### list empty page layout
* [x]  implementation
* [x]  write storybooks
* [x]  write tests

### extension details page layout
* [x]  implementation
* [x]  write storybooks
* [x]  write tests

### extension import page layout
* [x]  implementation
* [x]  write storybooks
* [x]  write tests

See syndesisio/syndesis#4043

gashcrumb:
This is like, done, isn't it?

elvisisking:
Yes. I have finished stories and jest tests.

PR #79, opened 2019-02-25
Work on developing the extension detail page and hooking it up with the list view

elvisisking:
- Work relating to #78 
- Initial implementation of the extension detail page
- See commit message for more details
- Forgot to mention in commit message that I added the Dashboard routes and resolvers files.


packages/api/src/WithExtension.tsx
```diff
@@ -0,0 +1,40 @@
...
+          actions: [],
+          extensionType: 'Steps',
+          name: '',
+          schemaVersion: '',
+        }}
+        initialValue={this.props.initialValue}
+      >
+        {({ read, response }) =>
+          this.props.children({
+            ...response,
```
riccardo-forina: unless you have to change the response somehow, I think you can just pass the `response` object as is

packages/ui/src/Customization/ExtensionDetail.css
```diff
@@ -0,0 +1,19 @@
...
+}
+
+.extension-title {
+  margin: 0;
+}
+
+/* .section-heading {
+} */
+
+button + button {
```
riccardo-forina: Namespace this please

packages/ui/src/Customization/ExtensionOverview.css
```diff
@@ -0,0 +1,8 @@
+.property-label {
```
riccardo-forina: namespace should be the same as the filename

packages/api/src/WithExtension.tsx
elvisisking: I'll remove `read`.

packages/ui/src/Customization/ExtensionDetail.css
elvisisking: Will do

packages/ui/src/Customization/ExtensionOverview.css
elvisisking: Will do

PR #80, opened 2019-02-25
feature(integrationDetailHistoryListView)

kahboom:
Opening this as a WIP/Draft Pull Request as I have some questions before I make this ready for review. Will also try to add the other sections of the Details tab if possible.

Changes:
- Add IntegrationDetailHistoryListView component and story.
- The history section, slightly confusingly, includes a row for Draft options, and then a series of rows for History items, see [this](https://syndesis-staging.b6ff.rh-idev.openshiftapps.com/integrations/i-LYmlhVFB6pKKaBQVSyez) example in staging. So, I have moved the Draft row of options to this here List View component and story (from the List View Item), as it made it easier to work with the data.

Questions:
- ~Is it possible to extract data for stories that would like to reference the same data item? For example, some integration properties (e.g. `version`, `updatedAt`) are used in multiple integration detail-related stories, so at the moment I just copy and paste them into each story.~

### DIFF
--- a/packages/ui/src/Integration/IntegrationDetailHistoryListView.tsx
+++ b/packages/ui/src/Integration/IntegrationDetailHistoryListView.tsx
@@ -14,43 +14,41 @@ export class IntegrationDetailHistoryListView extends React.Component<
 > {
   public render() {
     return (
-      <>
-        <Grid fluid={true} key={1}>
-          {this.props.integrationIsDraft ? (
-            <Grid.Row className="show-grid">
-              <Grid.Col xs={2} md={2}>
-                {<span>{this.props.i18nTextDraft}:</span>}
-              </Grid.Col>
-              <Grid.Col xs={10} md={10}>
-                <ListViewItem
-                  key={1}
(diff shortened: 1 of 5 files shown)


riccardo-forina:
> * Is it possible to extract data for stories that would like to reference the same data item? For example, some integration properties (e.g. `version`, `updatedAt`) are used in multiple integration detail-related stories, so at the moment I just copy and paste them into each story.

sure, we could add a `data` folder inside the `stories` one where commonly used data can just be placed in ready to be used.

kahboom:
>sure, we could add a `data` folder inside the `stories` one where commonly used data can just be placed in ready to be used.

Cool! I can do that in this PR or a separate one and create an issue for it, depending what you'd prefer. Let me know.

kahboom:
screenshot 2019-02-26 10 06 49

Getting incredibly annoyed with the space above the first list item, this didn't seem to be happening when I had this in the full detail page. I could use CSS to remove the margin-top, but probably not worth it, going to see if I can make any minor changes that would help.

PR #97, opened 2019-03-07
TTOOLS-588 Adds CreateVirtualization

mdrillin:
- add createVirtualization function to WithVirtualizationHelpers
- username prop was added to WithVirtualizationHelpers.  many of the repo paths require the username as part of the path.  
- modified the VirtualizationCreatePage to contain the form for entering a name and description.  Upon successful create, user is routed to list page.  (in future, this will route to details page instead)
- additions to i18n, plus other minor changes.

### DIFF
--- a/packages/api/src/WithVirtualizationHelpers.tsx
+++ b/packages/api/src/WithVirtualizationHelpers.tsx
@@ -1,12 +1,20 @@
+import { RestDataService } from '@syndesis/models';
 import * as React from 'react';
 import { ApiContext, IApiContext } from './ApiContext';
 import { callFetch } from './callFetch';
 
+const WORKSPACE_ROOT = '/tko:komodo/tko:workspace/';
+
 export interface IWithVirtualizationHelpersChildrenProps {
+  createVirtualization(
+    virtualizationName: string,
+    virtualizationDescription?: string
+  ): Promise<void>;
   deleteVirtualization(virtualizationName: string): Promise<void>;
 }
 
 export interface IWithVirtualizationHelpersProps {
+  username: string;
   children(props: IWithVirtualizationHelpersChildrenProps): any;
 }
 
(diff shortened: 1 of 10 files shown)


syndesis/src/modules/data/pages/VirtualizationCreatePage.tsx
```diff
@@ -1,21 +1,116 @@
...
+                    <>
+                      <Breadcrumb>
+                        <Link to={resolvers.dashboard.root()}>
+                          {t('shared:Home')}
+                        </Link>
+                        <Link to={resolvers.data.root()}>
+                          {t('shared:DataVirtualizations')}
+                        </Link>
+                        <span>
+                          {t('shared:Create')} {t('shared:DataVirtualization')}
```
elvisisking: You should reference one i18n key here. Then in the localization file reference other keys. That gives the language the choice to order them or not use them if they choose.

packages/api/src/WithVirtualizationHelpers.tsx
```diff
@@ -15,9 +23,41 @@ export class WithVirtualizationHelpersWrapped extends React.Component<
...
+  /**
+   * Creates a virtualization with the specified name and description
+   * @param virtName the name of the virtualization to create
+   * @param virtDesc the description of the virtualization to create
+   */
+  public async createVirtualization(
+    virtName: string,
+    virtDesc?: string
+  ): Promise<void> {
+    if (virtName) {
```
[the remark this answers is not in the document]
mdrillin: no, good point - I'll remove

syndesis/src/modules/data/pages/VirtualizationCreatePage.tsx
mdrillin: ok - I'll change that

riccardo-forina:
So it looks like ui storybook is broken. See the error here https://circleci.com/gh/syndesisio/syndesis-react-poc/275
Could you please fix it in a new PR?

PR #109, opened 2019-03-25
Implementation of Extension Import

elvisisking:
- work supporting issue #78 
- created shared DndFileChooser component to handle file uploads (uses react-dropzone library) and can be customized to accept just one file or multiple files.
- created DndFileChooser stories
- implemented ExtensionImportPage to work like the angular app
- created ExtensionImportReview and ExtensionImportCard components used by the import page (also created stories for these new components)
- changed `callFetch` to allow content type, accept, and referral policy to be set or removed
- changed `callFetch` to all a body that is not JSON


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react-poc/349/artifacts/0/home/circleci/project/doc/index.html)

packages/ui/src/Shared/DndFileChooser.tsx
```diff
@@ -0,0 +1,312 @@
...
+    this.state = {
+      allowMultiple: this.props.allowMultiple
+        ? this.props.allowMultiple
+        : false,
+      files: [],
+      notificationTimerDelay: this.props.notificationTimerDelay
+        ? this.props.notificationTimerDelay
+        : 4000,
+      notifications: [],
+    };
```
riccardo-forina: You are storing as state something that comes from the props, but you are not dealing with prop changes. This is a common anti-pattern. 

Check `Principle 1: Don’t Stop the Data Flow` here for more details https://overreacted.io/writing-resilient-components/#dont-stop-the-data-flow-in-rendering

packages/ui/src/Shared/DndFileChooser.tsx
```diff
@@ -0,0 +1,312 @@
...
+   * is set then `i18nUploadSuccessMessage` should not be set.
+   */
+  i18nUploadFailedMessage?: string;
+
+  /**
+   * A general, localized message for when a file upload was successful. This message will be shown
+   * along with an OK icon and appear next to the selected file message. If this property
+   * is set then `i18nUploadFailedMessage` should not be set.
+   */
+  i18nUploadSuccessMessage?: string;
```
riccardo-forina: I'm not sure about this; the comment says that `i18nUploadFailedMessage` and `i18nUploadSuccessMessage` are mutually exclusive, but `getUploadMessage` use both of these props depending on the state of the upload.

Shouldn't these two props be both required?

packages/ui/src/Shared/DndFileChooser.tsx
```diff
@@ -0,0 +1,312 @@
...
+                    }}
+                  />
+                </TimedToastNotification>
+              ))}
+            </ToastNotificationList>
+            <Grid
+              disabled={this.props.disableDropzone}
+              fluid={true}
+              className="dnd-file-chooser"
+              {...getRootProps({ refKey: 'dnd-file-chooser' })}
```
riccardo-forina: Is this custom `refKey` required?

packages/ui/src/Shared/DndFileChooser.tsx
elvisisking: The upload message uses the existence of those props to decide which icon to display next to the message. An alternative way would be to make them both required and add another property that indicates which message and icon to display. Is that what you had in mind?

packages/ui/src/Shared/DndFileChooser.tsx
elvisisking: I was thinking those 2 properties (allowMultiple, notificationTimerDelay) would not change. I was trying to add default values for those properties so that the parent didn't have to set them if they didn't want to. I can see your point where this might be confusing or maybe not behave the way someone might assume it does. Is there a better way to do this?

packages/ui/src/Shared/DndFileChooser.tsx
elvisisking: It is required by the react-dropzone component. I'm not sure what it is used for as you can set it to anything you want. Not setting it does not break the behavior but does cause errors to show in the console.

packages/ui/src/Shared/DndFileChooser.tsx
riccardo-forina: https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values

[This article](https://medium.com/@martin_hotell/react-typescript-and-defaultprops-dilemma-ca7f81c661c7) will give you an idea of the pitfalls of `defaultProps` with TS, but I wouldn't bother going that route, I'd just set the `defaultProps` static property and call it a day.

packages/ui/src/Shared/DndFileChooser.tsx
riccardo-forina: Ah, I see, but now I'm more confused. Shouldn't showing one message or the other depend on the state of the upload?

packages/ui/src/Shared/DndFileChooser.tsx
riccardo-forina: Alright, thanks!

packages/ui/src/Shared/DndFileChooser.tsx
elvisisking: Yes, it does depend on the state of the upload. The DND component accepts and rejects files based on if the file has the correct file extension and if multiple files are allowed. Once the DND component accepts or rejects the file the parent gets notified via an accept file or reject file callback. The parent can process the file(s) involved in those callbacks however they want. For instance, for extension import, when a file is accepted by the DND component, that file is sent to the backend. Based on that processing the parent can determine an upload state it wants displayed in the DND component. Does that make sense? How would you like to see this work?

packages/ui/src/Shared/DndFileChooser.tsx
elvisisking: I'll update and use the static `defaultProps` property.

packages/ui/src/Shared/DndFileChooser.tsx
riccardo-forina: Ah, I think I get it now. It kinda feels like this logic should be handled completely by the parent component, which could simply use some other components meant to handle successful/failed uploads.

I mean, it's just an `if` from the parent component standpoint, but I think it simplifies understanding how to use this component.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react-poc/351/artifacts/0/home/circleci/project/doc/index.html)

PR #111, opened 2019-03-29
feat(alertgroup): introduce pf4 alertgroup

seanforyou23:
This PR introduces PF4 AlertGroup component. It's dependent upon https://github.com/patternfly/patternfly-react/pull/1622 so you'll need to pair it with the [alert-group](https://github.com/seanforyou23/patternfly-react/tree/alert-group) branch of react-core from my pf fork to see it in action. Below are screenshots of three types of notifications being used within AlertGroup. PF3 ToastNotification is the one that's actually plugged into the Drag n Drop File Chooser (it's the only place we use notifications currently), just showing them all for illustration sake. Is there a way to hook into the server error scenario as seen in the linked issue below?

Housing PF3 Alerts
Screen Shot 2019-03-28 at 11 10 48 AM

Housing PF3 ToastNotification
Screen Shot 2019-03-28 at 8 33 53 PM

Housing PF4 Alerts (Sans PF4 Font)
Screen Shot 2019-03-28 at 11 16 14 AM

The error alerts for D&D File Chooser are currently set to 4 seconds, they are brief in nature so it makes sense, but then we have [this issue](https://github.com/syndesisio/syndesis-react-poc/issues/86) on the other end, which probably was trying to relay too much information via toast in the first place. Is there a rule of thumb we can follow to determine what's the best delay time to use?

### DIFF
--- a/packages/.storybook/addons.js
+++ b/packages/.storybook/addons.js
@@ -6,3 +6,4 @@ import '@storybook/addon-options/register';
 import '@storybook/addon-actions/register';
 import '@storybook/addon-links/register';
 import '@storybook/addon-viewport/register';
+import '@dump247/storybook-state/register';
--- a/packages/.storybook/vendor.css
+++ b/packages/.storybook/vendor.css
@@ -2,6 +2,10 @@
 @import url('~patternfly/dist/css/patternfly-additions.css');
 @import url('~patternfly-react/dist/css/patternfly-react.css');
 
+@import url('~@patternfly/patternfly/patternfly-variables.css');
+@import url('~@patternfly/patternfly/patternfly-shield-noninheritable.css');
+@import url('~@patternfly/patternfly/patternfly-common.css');
+
 html,
 body {
   height: 100%;
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -12,6 +12,10 @@
   "license": "MIT",
   "devDependencies": {
     "@babel/core": "^7.1.5",
+    "@dump247/storybook-state": "^1.5.2",
+    "@patternfly/patternfly": "^1.0.248",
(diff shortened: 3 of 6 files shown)


riccardo-forina:
So this can't really be reviewed until your other PR gets merged?

seanforyou23:
@riccardo-forina as I mentioned you can review this by pairing this branch in syndesis along with the one from pf-react which contains the corresponding changes that are needed. Posting this PR is in an effort to show progress and give anyone who wants/needs to test it a path forward.

riccardo-forina:
Gotcha, thanks.

I believe the notifications in the import extensions page were inline and not of the toast kind, which I think makes sense given that those alerts are more akin to validation errors than toast notifications. Did PF4 drop inline alerts for toast notifications?

Generically speaking, I think toast notifications should be handled centrally - by the app - and not directly by the single components, and offer an alternative way to read them again if needed ([I wrote a comment about this in the issue you linked](https://github.com/syndesisio/syndesis-react-poc/issues/86#issuecomment-477934401)).

elvisisking:
Your changes to `DndFileChooser` look great @seanforyou23. There are 2 other places (which I think we've talked about) that *might* benefit from your new alert group:
1. the [`DndFileChooser.getUploadMessage()`](https://github.com/syndesisio/syndesis-react-poc/blob/master/packages/ui/src/Shared/DndFileChooser.tsx#L162) method, and
1. the [`ExtensionImportCard.render()`](https://github.com/syndesisio/syndesis-react-poc/blob/master/packages/ui/src/Customization/ExtensionImportCard.tsx#L85) method which uses the inline alert as @riccardo-forina mentions.

These changes can be done in another PR.

PR #112, opened 2019-04-01
TTOOLS-601 - Added QueryResultsTable component and stories

blafond:


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react-poc/380/artifacts/0/home/circleci/project/doc/index.html)

blafond:
I like the idea of the GenericTable component, so I'll work on that.

The useMemo() is new to me (as most of this) and I'll adapt :)

1) The replacement QueryResultsTable.tsx didn't work out of the box. The 'rowKey' value isn't being set in the getTableColumns() useMemo() resulting in Table.Body warnings
2) I initialized it with an EMPTY state, and when I threw actual data at it, it failed with: "Warning: The final argument passed to useMemo changed size between renders. The order and size of this array must remain constant."   Note that the QueryResultsTable component is expected to be refreshed for multiple queries against a table that can contain different #'s of columns and rows.  So not sure how to tweak your new memoized logic?

Thx

blafond:
FYI.. after staring at it a little more, I see the rowKey value being set.

riccardo-forina:
Interesting, I don't see any warning using the stories that you added (I didn't touch them, just the implementation).

blafond:
The stories involve 3 separate static components with EMPTY query results and 2 sets of example data.  There is no re-setting of queryResults from the stories.

riccardo-forina:
Wait, why is this closed? I'm so sorry, I must have clicked on the wrong button!

blafond:
Working on a rewrite and additional components. Closing and will resubmit

PR #113, opened 2019-04-01
Added some Jest Tests for Customization Components

elvisisking:
- added dependency to `react-test-renderer` (used for jest snapshot testing)
- added comments to `IExtensionDetailProps`
- removed all uses of `data-testid`
- modified `ExtensionSupports` to only create a `Row` if there is content
- added 4 jest tests for customization components
- adding jest snapshots for the 4 jest tests

@riccardo-forina I'm getting lots of the following when I run `yarn test --watch`:
```
   console.error ../../node_modules/react-dom/cjs/react-dom.development.js:506
      Warning: This synthetic event is reused for performance reasons. If you're seeing this, you're accessing the property `view` on a released/nullified synthetic event. This is set to null. If you must keep the original synthetic event around, use event.persist(). See https://fb.me/react-event-pooling for more information.
```

Was hoping you could take a look. Also:
- I really thought the inline snapshots were really cluttering up the test file so I went with the generated snapshot file. Wdyt?
- I wanted to test that the delete confirmation dialog was being displayed when I hit the delete button. The only way I could figure out how to do that was to do that in `ExtensionDetail.spec.tsx` was to do `    expect(baseElement.classList).toContain('modal-open');` but doesn't *feel* like the right way to do it.


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react-poc/382/artifacts/0/home/circleci/project/doc/index.html)

packages/ui/tests/Customization/ExtensionDetail.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+    expect(snapshot).toMatchSnapshot();
+  });
+
+  it('Should open delete confirmation modal', () => {
+    const comp = <ExtensionDetail {...props} extensionUses={0} />;
+    const { baseElement, getByText } = render(comp);
+    const deleteButton = getByText(deleteLabel);
+    expect(deleteButton).not.toHaveAttribute('disabled'); // delete should be enabled
+    fireEvent.click(deleteButton); // open delete confirmation dialog
+    expect(baseElement.classList).toContain('modal-open');
```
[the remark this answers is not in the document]
elvisisking: Thanks! I changed a bit so that I could get the delete button on the confirmation, click it, and make sure the delete callback was called.

packages/ui/tests/Customization/ExtensionDetail.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+    // supports section title
+    expect(queryAllByText(supportedStepsText)).toHaveLength(1);
+
+    // usage section title
+    expect(queryAllByText(usageLabel)).toHaveLength(1);
+
+    // test snapshot
+    const snapshot = renderer.create(comp).toJSON();
+    expect(snapshot).toMatchSnapshot();
```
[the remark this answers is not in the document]
elvisisking: Added another commit that removes `react-test-renderer` and removes all the snapshot tests that create snapshot files.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react-poc/384/artifacts/0/home/circleci/project/doc/index.html)

packages/ui/tests/Customization/ExtensionDetail.spec.tsx
```diff
@@ -79,22 +78,27 @@ export default describe('ExtensionDetail', () => {
...
-    // test snapshot
-    const snapshot = renderer.create(comp).toJSON();
-    expect(snapshot).toMatchSnapshot();
+    // click the delete button so that the delete confirmation dialog opens
+    fireEvent.click(deleteButton);
+
+    // wait for the delete dialog to show by looking for the delete button
+    const elements = await waitForElement(() => {
+      // find the delete button
+      return baseElement.getElementsByClassName('btn btn-danger');
```
riccardo-forina: You are testing the implementation this way. Why not querying for the `deleteLabel` text?

elvisisking: Tried that but ran into the problem that there are 2 components in the DOM that have that text (the delete button in the popup confirmation dialog, and the delete button on the details component).

riccardo-forina: Ah, gotcha. And since the dialog is just a wrapper over the one exposed by PF we can't add a `data-testid` to its button either. 

Maybe wrap the overlay in some div with a data-testid, query for that and then use getByText on that element? Something like this (warning: I haven't tested it lol):

```
import {getByText, ... } from 'react-testing-library';

...

const {getByTestId, ...} = render(...);
const dialog = getByTestId('dialog');
expect(getByText(dialog, deleteLabel)).toBeInTheDocument();
```

elvisisking: Isn't that kind of knowing the implementation? Why would this way be better than what I originally had? (only saying this bc I'm having trouble getting your idea to work:)

riccardo-forina: It's a fair question. It's not a problem knowing about the implementation, actually, you have to know how the thing works to test it, right? It's testing the implementation that's the problem. The way you have will break if another "danger" button gets added to the dialog, where with the suggested change the test will continue to pass.

I understand that my example is an unlikely scenario, but it's important to remember that a component should be tested from the user standpoint. A CSS class is invisible to a user, so it's better not to rely on it unless strictly necessary.

> On top of the queries provided by the testing library, you can using the regular querySelector DOM API to query elements. Note that using this as an escape hatch to query by class or id is a bad practice because users can't see or identify these attributes. Use a testid if you have to.

https://testing-library.com/docs/guide-which-query#manual-queries

elvisisking: Modified test as we discussed @riccardo-forina.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react-poc/386/artifacts/0/home/circleci/project/doc/index.html)

PR #117, opened 2019-04-08
feat: adds some error resilience to the UI

riccardo-forina:
This adds a way to catch errors when rendering integration items, and shows a fallback instead. 

It is in no way to be considered the best way to deal with this kind of problem, it's just a patch; that said, I think the approach is interesting and we could try it out and in case, extend it/make it better.

## Changes

Some bogus integrations.
Screenshot 2019-04-08 at 20 46 28

What is displayed clicking on the "Integration JSON" button.
Screenshot 2019-04-08 at 20 46 34


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react-poc/460/artifacts/0/home/circleci/project/doc/index.html)

PR #140, opened 2019-04-12
[DO NOT MERGE] test commit that uses pf4 Alert component

riccardo-forina:
Just leveraging CI to see if this generally works.

### DIFF
--- a/app/ui-react/packages/ui/package.json
+++ b/app/ui-react/packages/ui/package.json
@@ -72,6 +72,7 @@
     "react-router-dom": "^4.3.1"
   },
   "dependencies": {
+    "@patternfly/react-core": "^2.10.9",
     "classnames": "^2.2.6",
     "history": "^4.7.2",
     "react-dropzone": "^10.0.0",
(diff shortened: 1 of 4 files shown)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/603/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

seanforyou23:
Update: While this PR built for me just fine, it still seems the integration with PF4 is still pretty fragile. I don't understand why there needs to be two occurrences of `@patternfly/react-core` in the package.json. In addition, after installing `@patternfly/patternfly` I found myself looking at the same error as I was seeing previously (below)

Screen Shot 2019-04-17 at 10 16 54 AM

seanforyou23:
Another update: @riccardo-forina found that by adding `"**/rxjs"` to the nohoist array in the root package.json, the problems go away.

riccardo-forina:
This should not be needed anymore

PR #177, opened 2019-04-24
feat: decouple auto-form from syndesis model

### LINKED ISSUE #176: Decouple auto-form from syndesis model
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description
I'm gonna copy over `ConfigurationProperty` and related definitions to autoform and make an equivalent, that way we can get rid of all the optional fields (and OptionalInt) we don't care about and have required feilds so it's easier to use.  Then worst case we can add a converter function to the app to satisfy typescript and handle any future model changes.

gashcrumb:
fixes #176

Also adds some groundwork for other auto-form improvements and a bunch of simpler, easier to look at storybooks.

### DIFF
--- a/app/ui-react/packages/auto-form/package.json
+++ b/app/ui-react/packages/auto-form/package.json
@@ -78,7 +78,6 @@
     }
   },
   "dependencies": {
-    "@syndesis/models": "*",
     "@syndesis/ui": "*",
     "formik": "^1.5.1",
     "patternfly-react": "^2.29.15"
--- a/app/ui-react/packages/auto-form/src/models.ts
+++ b/app/ui-react/packages/auto-form/src/models.ts
@@ -55,7 +55,6 @@ export interface IFormDefinitionProperty {
   enum?: IFormPropertyValue[];
   generator?: string;
   placeholder?: string;
-  javaType?: string;
   connectorValue?: string;
   relation?: IFormPropertyRelation[];
   controlHint?: string;
--- a/app/ui-react/packages/auto-form/src/models.ts
+++ b/app/ui-react/packages/auto-form/src/models.ts
@@ -1,9 +1,9 @@
-import { IConfigurationProperty } from '@syndesis/models';
 import { FieldProps } from 'formik';
 
 export interface IFormDefinition {
-  [name: string]: IConfigurationProperty;
+  [name: string]: IFormDefinitionProperty;
 }
+
 export interface IFormValue {
   [name: string]: any;
 }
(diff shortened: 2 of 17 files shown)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/783/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

gashcrumb:
Snuck some more work in on this one, but it's ready for review

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/788/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/packages/auto-form/package.json
```diff
@@ -78,7 +78,6 @@
     }
   },
   "dependencies": {
-    "@syndesis/models": "*",
```
riccardo-forina: 🎉

app/ui-react/packages/auto-form/src/models.ts
```diff
@@ -21,12 +21,52 @@ export interface IFormField {
...
+export interface IFormPropertyValue {
+  value?: string;
+  label?: string;
+}
+
+export interface IFormFieldAttributes {
+  [name: string]: any;
+}
+
+export interface IFormDefinitionProperty {
```
riccardo-forina: I think this interface could see a good many of the props dropped? Eg. the `javaType`.
Not needed for this PR, mind, just something we could try to do eventually.

app/ui-react/packages/auto-form/src/widgets/FormDurationComponent.tsx
```diff
@@ -0,0 +1,147 @@
...
+  IFormControl,
+  IFormDurationComponentState
+> {
+  private inputField: HTMLInputElement = undefined as any;
+  constructor(props: IFormControl) {
+    super(props);
+    const index =
+      durations.findIndex(d => !(this.props.field.value / d.value >= 1)) - 1;
+    const duration = durations[index] || durations[0];
+    const value = this.calculateDuration(duration, this.props.field.value);
```
riccardo-forina: I think this value copy leads to stale data when the component receives a new value since there is no `onComponentDidUpdate` handler. 

I don't think you want to have a copy of the value in the state at all. The calculation could be done in the render method, so to always use the value passed in the props. Duration is ok to be kept in the state since it only changes the way the value is displayed to the user.

app/ui-react/packages/auto-form/src/models.ts
gashcrumb: Let me see about pulling `javaType` specifically.  Some of the other `group` or `label` maybe could be useful at some point if we want to support organizing form controls into groups, and I have to double-check `connectorValue`, `tags`, and `kind`.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/796/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #185, opened 2019-04-25
fix: force a fetch after performing a server call for environments

### LINKED ISSUE #183: Do a fetch after deleting/renaming items in the CI/CD list
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description
The ci/cd page only does polling, so deleting, renaming and adding things can take a few seconds before the results appear on the page.  The WithEnvironments component should probably also expose a function to enable it's consumer to call read() whenever it's needed.

gashcrumb:
fixes #183

@mdrillin FYI, I extended IFetchState with my type and added the read operation, then I'm calling that read operation when the call completes, this will then trigger re-rendering the list.  Hope that helps!

### DIFF
--- a/app/ui-react/packages/api/src/WithEnvironments.tsx
+++ b/app/ui-react/packages/api/src/WithEnvironments.tsx
@@ -3,9 +3,13 @@ import { IFetchState } from './Fetch';
 import { SyndesisFetch } from './SyndesisFetch';
 import { WithPolling } from './WithPolling';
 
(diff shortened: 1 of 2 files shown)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/831/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #195, opened 2019-04-29
feat: show steps based on the integration and position; switch to flow id instead of idx

riccardo-forina:
Ported the logic from the Angular codebase, steps should now be filtered based on the current integration object and the position.

Also switched to flow ids instead of the positional index, to align to how the functions work.


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/895/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #207, opened 2019-04-30
fix: hide the results bar when there's no filters active

### LINKED ISSUE #181: Remove "# Results" from toolbar when at default view
## This is a...

<pre><code>
[ ] Feature request
[ ] Bug report  
</code></pre>

## Description
Currently, the "# Results" is a constant element on the page, which means it shows up in both the default view and when the filter/sort is actually active. 

Should remove the "# Results" when filter/sort is in use. Only when users start interacting with filter/sort then we want to show the results. 

See screenshots below: 
Screen Shot 2019-04-25 at 2 51 49 PM

Screen Shot 2019-04-25 at 2 51 57 PM

gashcrumb:
fixes #181


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/933/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #163, opened 2019-04-19
feat: go on different routes when clicking a connection|step creating an integration

riccardo-forina:
So this is a big one.

Before this PR the editor was composed by many, many pages, which were basically always the same set of pages with subtle changes to reflect the place where they were plugged-in. This was because of the editor having many ways to ultimately do the same thing, which is to create/edit an integration:

* create: start connection - this is triggered when you start creating an integration from scratch
* create: finish connection - same as above, after you completed the start connection bit
* create: edit - the "true" editor, where you can add/remove/configure steps at leisure
* integration: edit - same as above, but for an existing integration

Not listed above are all the sub-pages required to handle all the different scenarios the editor supports: endpoints, steps, data mapper, api provider, template, possibly others I don't remember.

To support all these sub-pages without this refactoring, we would have had to implement each sub-page 4 times, one for every major entry point described above.
To put it mildly, this was a mess.

Now the sub-pages are pluggable components, each requiring as props the links to connect each sub-page to its following logical one. This connection is done on the IntegrationCreatorApp and IntegrationEditorApp components, together with the UI customizations required to reflect the entry point the page is plugged into.

This called for a heavy refactoring of the resolvers and of the routes. The resolvers, in particular, can now optionally take additional information about the route they are dealing with to ensure that the connection route-params-state is typesafe, by pulling the params/state interfaces used by the pages and linking them to the route (as in, the URL) interface.

PR #209, opened 2019-05-01
chore: editor placeholder pages

riccardo-forina:
This introduces the `EditorApp` component, a pluggable "app" that will handle the routes required to handle adding/editing a step to an integration.

Its main goal is to make sure that regardless of where the app has been plugged, all the required routes and sub-components will be set. This together with the previous PR #163 should remove the risk of extending the editor with new features forgetting to do it everywhere is needed.

There are a lot of `TODO` placeholders for the pages right now, but the basic routing and wiring should be in place. Clicking on any step object when adding a step to an integration should never give an error, but redirect to the right URL and show a "TODO" as the page content.

I'm not super happy about how the whole thing looks right now, there is _a lot_ of code that needs to be written to wire a page to the editor "only" for being able to do this in a typesafe way. I hope this will pay in the long run.

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/SelectConnectionPage.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/SelectConnectionPage.tsx
@@ -25,16 +25,15 @@ import {
 import { WithLoader, WithRouteData } from '@syndesis/utils';
 import * as H from 'history';
 import * as React from 'react';
-import { Omit } from 'react-router';
 import { ApiError, PageTitle } from '../../../../shared';
 import {
   ISelectConnectionRouteParams,
   ISelectConnectionRouteState,
 } from './interfaces';
 
-export interface IUIStep extends Omit<StepKind, 'stepKind'> {
+export interface IUIStep extends StepKind {
   icon: string;
-  stepKind: 'api-provider' | StepKind['stepKind'];
+  uiStepKind: 'api-provider' | StepKind['stepKind'];
 }
 
 export function toStepKindCollection(
@@ -50,7 +49,7 @@ export function toStepKindCollection(
           description: c.description || '',
           icon: getConnectionIcon(process.env.PUBLIC_URL, c),
           properties: undefined,
-          stepKind:
+          uiStepKind:
             c.connectorId === 'api-provider' ? 'api-provider' : 'endpoint',
         } as IUIStep)
     ),
@@ -79,6 +78,7 @@ export function toStepKindCollection(
               name: a.name,
               properties,
               stepKind: 'extension',
+              uiStepKind: 'extension',
             });
           }
         });
@@ -89,10 +89,10 @@ export function toStepKindCollection(
     ...steps.map(s => ({
       ...s,
       icon: `${process.env.PUBLIC_URL}${getStepKindIcon(s.stepKind)}`,
-      stepKind: s.stepKind,
+      uiStepKind: s.stepKind,
     })),
   ]
-    .filter(s => !!s.stepKind) // this should never happen
+    .filter(s => !!s.uiStepKind) // this should never happen
     .sort((a, b) => a.name.localeCompare(b.name));
 }
 
@@ -144,7 +144,7 @@ export class SelectConnectionPage extends React.Component<
           const { integration = getEmptyIntegration() } = state;
           const positionAsNumber = parseInt(position, 10) || 0;
           const getStepHref = (step: IUIStep) => {
-            switch (step.stepKind) {
+            switch (step.uiStepKind) {
               case 'api-provider':
                 return this.props.apiProviderHref(step, params, state);
               case 'endpoint':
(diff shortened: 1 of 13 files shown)


riccardo-forina:
This is a sample usage of the `EditorApp` component. It requires the route and the relative children component that will be rendered on a route match, for every "sub-editor" we support; right now the list consists of endpoint (aka connection), api provider, template, data mapper, filter, template, step (the hard-coded ones like the log), and extension.

```tsx
<EditorApp
  selectStepPath={routes.create.start.selectStep}
  selectStepChildren={startStepSelectConnectionPage}
  endpointEditor={{
    selectActionPath: routes.create.start.connection.selectAction,
    selectActionChildren: startStepSelectActionPage,
    configureActionPath:
      routes.create.start.connection.configureAction,
    configureActionChildren: startStepConfigureActionPage,
    describeDataPath: routes.create.start.connection.describeData,
    describeDataChildren: TODO,
  }}
  apiProvider={{
    uploadPath: routes.create.start.apiProvider.upload,
    uploadChildren: <UploadPage />,
    reviewPath: routes.create.start.apiProvider.review,
    reviewChildren: <ReviewPage />,
    editPath: routes.create.start.apiProvider.edit,
    editChildren: <EditPage />,
  }}
  template={{
    templatePath: routes.create.start.template,
    templateChildren: TODO,
  }}
  dataMapper={{
    mapperPath: routes.create.start.dataMapper,
    mapperChildren: TODO,
  }}
  basicFilter={{
    filterPath: routes.create.start.basicFilter,
    filterChildren: TODO,
  }}
  step={{
    configurePath: routes.create.start.step,
    configureChildren: TODO,
  }}
  extension={{
    configurePath: routes.create.start.extension,
    configureChildren: TODO,
  }}
/>
```

The `selectStepChildren` element, which must be of type `SelectConnectionPage`, will handle the actual routing between the sub-editors. Each step - as in the step element shown in the UI - points to an href that depends on the underlying `stepKind` property. The code that handles that looks like this:

```ts
const getStepHref = (step: IUIStep) => {
  switch (step.uiStepKind) {
    case 'api-provider':
      return this.props.apiProviderHref(step, params, state);
    case 'endpoint':
    case 'connector':
      return this.props.connectionHref(
        step as ConnectionOverview,
        params,
        state
      );
    case 'expressionFilter':
    case 'ruleFilter':
      return this.props.filterHref(step, params, state);
    case 'extension':
      return this.props.extensionHref(step, params, state);
    case 'mapper':
      return this.props.mapperHref(step, params, state);
    case 'headers':
      throw new Error(`Can't handle stepKind ${step.stepKind}`);
    case 'template':
      return this.props.templateHref(step, params, state);
    case 'choice':
    case 'split':
    case 'aggregate':
    case 'log':
    default:
      return this.props.stepHref(step, params, state);
  }
};
```

This is sample usage of the `SelectConnectionPage` component, as used in the "New integration: select start connection" step:

```tsx
<SelectConnectionPage
  apiProviderHref={resolvers.create.start.apiProvider.upload}
  connectionHref={(connection, params, state) =>
    resolvers.create.start.connection.selectAction({
      connection,
      ...params,
      ...state,
    })
  }
  filterHref={resolvers.create.start.basicFilter}
  extensionHref={resolvers.create.start.extension}
  mapperHref={resolvers.create.start.dataMapper}
  templateHref={resolvers.create.start.template}
  stepHref={resolvers.create.start.step}
  {/* ...other props... */}
/>
```

riccardo-forina:
The above snippet of codes are repeated for all the points in the integration editor app were the "EditorApp" is plugged, so:

* New integration: select start connection
* New integration: select finish connection
* New integration: add step
* New integration: edit step (without the possibility to change the step type)
* Edit existing integration: add step
* Edit existing integration: edit step (without the possibility to change the step type)

app/ui-react/syndesis/src/modules/integrations/components/editor/SelectConnectionPage.tsx
```diff
@@ -135,27 +139,38 @@ export class SelectConnectionPage extends React.Component<
...
-                break;
+              case 'expressionFilter':
+              case 'ruleFilter':
+                return this.props.filterHref(step, params, state);
+              case 'extension':
+                return this.props.extensionHref(step, params, state);
+              case 'mapper':
+                return this.props.mapperHref(step, params, state);
+              case 'headers':
+                throw new Error(`Can't handle stepKind ${step.stepKind}`);
```
riccardo-forina: @gashcrumb do you have any idea about this stepKind?

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/943/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/integrations/components/editor/SelectConnectionPage.tsx
gashcrumb: Hmm, that's a new one.  I wonder if it's one of the extensions there on staging.  Maybe we need a default case that just sends those to the regular step configuration page that just uses autoform.

gashcrumb:
Poking around, I hit this often when editing an integration on staging and adding a step:

image

For example I picked 'test3' I think and tried to add a 'log' connection to it.  seems the integration object at this spot isn't defined:

image

I did ctrl-F5 and opened the dev tools so I'm pretty sure my cache is clear.  I'll keep poking around though too :-)

riccardo-forina:
FYI the problem is the placeholder resolvers for all the routes, that right now simply return a string and don't set the params/state required for the pages to work.

PR #229, opened 2019-05-04
fix: fix: CircleCI `publish` image/tag for push

zregvart:
Seems that the CircleCI Orb commands are independently configured so we
need to repeat the image/tag parameters.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,7 +83,8 @@ jobs:
           extra_build_args: '--ulimit nofile=1024'
           image: syndesis/syndesis-ui
           tag: latest-react
-      - docker-publish/deploy
+      - docker-publish/deploy:
+        image: syndesis/syndesis-ui:latest-react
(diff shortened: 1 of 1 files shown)


riccardo-forina:
Uhm it looks like it doesn’t build

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1084/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

zregvart:
Yeah I was missing a colon and the `tag` parameter isn't allowed for `deploy` command, fixed and force pushed the fix [looks good now](https://circleci.com/gh/syndesisio/syndesis-react/1083).

I'd like to simplify it even further by using the `publish` job instead of `check`, `build`, `deploy` commands. Can't figure out the syntax might to a new PR to try to clean that up.

PR #233, opened 2019-05-04
fix: ci docker publish, specify Dockerfile path

zregvart:
Turns out that I should read the source of the Docker Orb a bit better,
it is not sufficient to just specify the `path`, `dockerfile` needs to
be specified also.


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1111/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #236, opened 2019-05-05
fix: dedupe bulletin board messages and combine details as needed

gashcrumb:
Instead of showing many alerts of the same type this PR combines them and merges the 'detail' field, so for example this one had tons of the same alert, now it's:

image

@dongniwang FYI found a nice solution for this nagging problem :-)

For reference, this is what this same integration looked like before:
image

and in the old app just for funs:
image


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1130/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

Issue #243, opened 2019-05-07
Help Entries


seanforyou23:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

For the help dropdown in the primary header, we are missing the following dropdown items

[ ] Sample Integration Tutorials
[ ] User Guide
[ ] Connectors Guide
[ ] Support
[ ] Contact Us

And the user dropdown's logout link doesn't currently work.

Let's wire these up.

PR #245, opened 2019-05-07
feat(help-dropdown): add help dropdown entries

seanforyou23:
This PR adds items to the help dropdown menu.

Screen Shot 2019-05-07 at 2 47 57 PM

Should help resolve https://github.com/syndesisio/syndesis-react/issues/243

TODO: Wire up logout for the user dropdown

### DIFF
--- a/app/ui-react/packages/ui/src/Shared/HelpDropdown.tsx
+++ b/app/ui-react/packages/ui/src/Shared/HelpDropdown.tsx
@@ -12,12 +12,12 @@ import * as React from 'react';
 export interface IHelpDropdownProps {
   className?: string;
   isOpen: boolean;
-  launchAboutModal: any;
-  launchSupportPage: any;
-  launchSampleIntegrationTutorials: any;
-  launchUserGuide: any;
-  launchConnectorsGuide: any;
-  launchContactUs: any;
+  launchAboutModal: () => void;
+  launchSupportPage: () => void;
+  launchSampleIntegrationTutorials: () => void;
+  launchUserGuide: () => void;
+  launchConnectorsGuide: () => void;
+  launchContactUs: () => void;
 }
 
 export interface IHelpDropdownState {
@@ -28,8 +28,8 @@ export class HelpDropdown extends React.Component<
   IHelpDropdownProps,
   IHelpDropdownState
 > {
-  public onToggle: any;
-  public onSelect: any;
+  public onToggle: (isOpen: boolean) => void;
+  public onSelect: () => void;
 
   constructor(props: IHelpDropdownProps) {
     super(props);
@@ -41,7 +41,7 @@ export class HelpDropdown extends React.Component<
         isOpen,
       });
     };
-    this.onSelect = (event: MouseEvent) => {
+    this.onSelect = () => {
       this.setState({
         isOpen: !this.state.isOpen,
       });
--- a/app/ui-react/packages/ui/src/Shared/HelpDropdown.tsx
+++ b/app/ui-react/packages/ui/src/Shared/HelpDropdown.tsx
@@ -13,6 +13,11 @@ export interface IHelpDropdownProps {
   className?: string;
   isOpen: boolean;
   launchAboutModal: any;
+  launchSupportPage: any;
+  launchSampleIntegrationTutorials: any;
+  launchUserGuide: any;
+  launchConnectorsGuide: any;
+  launchContactUs: any;
 }
 
 export interface IHelpDropdownState {
@@ -45,14 +50,39 @@ export class HelpDropdown extends React.Component<
 
   public render() {
     const { isOpen } = this.state;
+    const {
+      launchSampleIntegrationTutorials,
+      launchUserGuide,
+      launchConnectorsGuide,
+      launchSupportPage,
+      launchContactUs,
+      launchAboutModal,
+    } = this.props;
     const dropdownItems = [
       <DropdownItem
-        key="action"
+        key="sampleIntegrationTutorials"
         component="span"
-        onClick={() => {
-          this.props.launchAboutModal();
-        }}
+        onClick={launchSampleIntegrationTutorials}
       >
+        Sample Integration Tutorials
+      </DropdownItem>,
+      <DropdownItem key="userGuide" component="span" onClick={launchUserGuide}>
+        User Guide
+      </DropdownItem>,
+      <DropdownItem
+        key="connectorsGuide"
+        component="span"
+        onClick={launchConnectorsGuide}
+      >
+        Connectors Guide
+      </DropdownItem>,
+      <DropdownItem key="support" component="span" onClick={launchSupportPage}>
+        Support
+      </DropdownItem>,
+      <DropdownItem key="contactUs" component="span" onClick={launchContactUs}>
+        Contact Us
+      </DropdownItem>,
+      <DropdownItem key="action" component="span" onClick={launchAboutModal}>
         About
       </DropdownItem>,
     ];
--- a/app/ui-react/syndesis/src/app/UI.tsx
+++ b/app/ui-react/syndesis/src/app/UI.tsx
@@ -99,88 +99,118 @@ export const UI: React.FunctionComponent<IAppUIProps> = ({ routes }) => {
                 notificationTimerDelay={8000}
                 removeNotificationAction={onRemoveNotification}
               />
-              <AppLayout
-                onShowAboutModal={toggleAboutModal}
-                appNav={
-                  <AppTopMenu username={'developer'}>
-                    <PfDropdownItem>
-                      <Link
-                        to={'/logout'}
-                        className="pf-c-dropdown__menu-item"
-                        children={t('Logout')}
-                      />
-                    </PfDropdownItem>
-                  </AppTopMenu>
-                }
-                verticalNav={routes.map((route, index) =>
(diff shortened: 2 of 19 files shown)


riccardo-forina:
> is there a support route available? If not, should I create one? Or should I omit that item from the helpdropdown for now?

No there is not, I think we can expand this PR to include that too if you are ok with it.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1171/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

seanforyou23:
Alrighty, I've pushed some updates. Almost there, but there's an issue with the routing. Two things;

1) I can't seem to get the SupportPage component to render when hitting the support route unless it's registered here - https://github.com/syndesisio/syndesis-react/pull/245/files#diff-97c6800a2f8a73ad518dae66d5d1e87cR113 - is this expected? We don't have a Support item in the angular UI, so was hoping to avoid having to add it in the React UI.

2) When you select the support route, the url doubles up like `/support/support`. Probably something simple, but I haven't figured out why this happens yet.

Maybe these issues are related? In any case, feel free to provide feedback around the rest of the implementation. Thx!

app/ui-react/syndesis/src/modules/support/routes.ts
```diff
@@ -0,0 +1,8 @@
+import { include } from 'named-urls';
+
+export default include('/support', {
+  root: '',
+  support: include('support', {
+    root: '',
+  }),
```
riccardo-forina: You can remove all but the first root, and maybe rename it to support

app/ui-react/syndesis/src/index.tsx
```diff
@@ -108,6 +109,11 @@ ReactDOM.render(
                                 label: 'Settings',
                                 to: routes.settings.root,
                               } as IAppRoute,
+                              {
+                                component: SupportModule,
+                                label: 'Support',
+                                to: routes.support.root,
+                              } as IAppRoute,
```
riccardo-forina: Yeah it will render, but don’t worry. This needs to be cleaned up in a later moment, we can address this too then.

app/ui-react/packages/ui/src/Shared/HelpDropdown.tsx
```diff
@@ -46,6 +47,61 @@ export class HelpDropdown extends React.Component<
   public render() {
     const { isOpen } = this.state;
     const dropdownItems = [
+      <DropdownItem
+        key="sampleIntegrationTutorials"
+        component="span"
+        onClick={() => {
+          window.open(
+            'https://access.redhat.com/documentation/en-us/red_hat_fuse/7.3/html-single/fuse_online_sample_integration_tutorials/',
```
riccardo-forina: I think all these click handlers should be exposed to the app. These links shouldn’t be hard coded, not in the ui anyway.

app/ui-react/packages/ui/src/Shared/HelpDropdown.tsx
```diff
@@ -46,15 +50,20 @@ export class HelpDropdown extends React.Component<
...
+      launchConnectorsGuide,
+      launchSupportPage,
+      launchContactUs,
+      launchAboutModal,
+    } = this.props;
     const dropdownItems = [
       <DropdownItem
         key="sampleIntegrationTutorials"
         component="span"
         onClick={() => {
```
riccardo-forina: No need for the fat arrow here, you can pass the handler directly

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1190/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/packages/ui/src/Shared/HelpDropdown.tsx
```diff
@@ -13,6 +13,11 @@ export interface IHelpDropdownProps {
   className?: string;
   isOpen: boolean;
   launchAboutModal: any;
+  launchSupportPage: any;
```
riccardo-forina: If I were to be nitpicky I'd say that instead of any you should have written`(event: React.MouseEvent<HTMLAnchorElement>) => void | false`, where `React.MouseEvent` is the generic type that React uses for mouse related events - a click in this case, `HTMLAnchorElement` is [the underlying type of the `DropdownItem`](https://github.com/patternfly/patternfly-react/blob/master/packages/patternfly-4/react-core/src/components/Dropdown/DropdownItem.d.ts#L3), and the return `void | false` is to describe the allowed return values accordingly to the [event handler processing algorithm](https://html.spec.whatwg.org/multipage/webappapis.html#the-event-handler-processing-algorithm)  for this kind of event.

But this would be overkill. I'd simply write it as a `() => void` since that's what our component expects to handler to do.

app/ui-react/syndesis/src/modules/support/pages/SupportPage.tsx
```diff
@@ -0,0 +1,19 @@
+import { PageSection, SupportHeader } from '@syndesis/ui';
+import * as React from 'react';
+import { PageTitle } from '../../../shared';
+
+export function SupportPage() {
+  return (
+    <>
+      <PageTitle title="Support" />
+      <SupportHeader
```
riccardo-forina: We have a component to handle headers in the app, it's the `SimplePageHeader`. I'm replacing "custom" headers with it every time I can to be sure that we render the header always in the right way, in terms of text sizing and structure.

app/ui-react/syndesis/src/modules/support/pages/SupportPage.tsx
```diff
@@ -0,0 +1,19 @@
...
+        i18nPageTitle={'Support'}
+        i18nDescription={`To obtain support, download diagnostic information through this page and open a request on the <a href="https://access.redhat.com/support/cases/#/case/new">Red Hat Customer portal</a>. If you have any issues please see the support <a href="https://access.redhat.com/solutions/2112">instructions</a>.`}
+      />
+
+      <PageSection>
+        <h2>TODO</h2>
```
riccardo-forina: Currently, we show the same content you added to the about dialog plus a form to download some extra diagnostic. I think we should implement it as it is already since it's all info that can be handy when opening a ticket.

seanforyou23: @riccardo-forina I've added the basic version information and addressed the other bits of feedback. I think we're at a good point in time to merge what we have now and can flesh out the rest of the support page in a follow PR. I'd also like to work in the WithLoader support since there are api calls involved here, and don't want to drag this PR out for all that. Do you think what I've added so far enough for now?

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1230/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/app/UI.tsx
```diff
@@ -99,88 +99,118 @@ export const UI: React.FunctionComponent<IAppUIProps> = ({ routes }) => {
...
-                    >
-                      {(route as IAppRouteWithChildrens).childrens.map(
-                        (subRoute, subIndex) => (
+              <WithRouter>
+                {({ history, match }) => {
+                  return (
+                    <AppLayout
+                      onShowAboutModal={toggleAboutModal}
+                      onSelectSupport={() => {
+                        history.push('support');
```
riccardo-forina: Erh... sorry I didn't spot this earlier. You should use a resolver to retrieve the right URL, not a hardcoded string. Something like `resolvers.support.root()`

app/ui-react/packages/ui/src/Shared/HelpDropdown.tsx
```diff
@@ -13,6 +13,11 @@ export interface IHelpDropdownProps {
...
+  launchSampleIntegrationTutorials: any;
+  launchUserGuide: any;
+  launchConnectorsGuide: any;
+  launchContactUs: any;
```
riccardo-forina: So you decided to keep the `any` type?

app/ui-react/syndesis/src/modules/support/pages/SupportPage.tsx
```diff
@@ -0,0 +1,36 @@
+import { WithApiVersion } from '@syndesis/api';
+import { PageSection, SimplePageHeader, SupportPageBody } from '@syndesis/ui';
+import * as React from 'react';
+import { PageTitle } from '../../../shared';
+
+export function SupportPage() {
```
riccardo-forina: `export const SupportPage: React.FunctionComponent = () => {` would be more explicit and in line with the rest of the code.

app/ui-react/packages/ui/src/Shared/HelpDropdown.tsx
seanforyou23: Nope, just missed these!

seanforyou23: Just realized I hadn't pressed the save button after making the changes the first time 🤦‍♂️

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1232/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #290, opened 2019-05-15
feat: re-fetch upon receiving connected message; clear error on read

### LINKED ISSUE #119: When WithServerEvents hits a problem, it's bad
Staging started returning a 502 which caused the app to just show this:

image

and

image

When we have notification support this should cause the http error notification, for reference the angular app [does it this way](https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/error-handler/http-error.interceptor.ts#L77).

gashcrumb:
fixes #119 

I was totally gonna sneak this onto my last PR but @riccardo-forina merged it while I had stepped away :-)

Anyway, this fully fixes #119, I noticed that the app wasn't reverting back to showing the items once successfully reconnecting.  This ensures that when we reconnect and receive a connected event from the server the app will re-read, and that read will clear any previous read errors.  So if you don't happen to refresh the page it'll just go back to showing whatever list you were looking at.

### DIFF
--- a/app/ui-react/packages/api/src/Fetch.tsx
+++ b/app/ui-react/packages/api/src/Fetch.tsx
@@ -66,7 +66,7 @@ export class Fetch<T> extends React.Component<IFetchProps<T>, IFetchState<T>> {
 
   public async read() {
     try {
-      this.setState({ loading: true });
+      this.setState({ error: false, errorMessage: undefined, loading: true });
       const response = await callFetch({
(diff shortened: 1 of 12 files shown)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1412/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #302, opened 2019-05-16
Delete Step in Integration Editor

### LINKED ISSUE #259: Editor - implement the delete step action and confirmation dialog
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description
The trash icon needs to be hooked up for steps.  In all cases the trash icon should prompt the user if they really want to remove the step.  

For steps at the start and end of the flow the step should be removed but the user should then be brought to the step select page for that position and be required to set a step/connection and configure it.

For steps in the middle positions after confirmation the step is simply removed from the step array.

Special steps such as the API provider start/finish connections and the conditional flow start/finish connections cannot be removed, so the trash icon should be either hidden or disabled.  The angular app hides them.

kahboom:
Delete steps in the Edit Integration page.

fix #259

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/components/IntegrationEditorStepAdder.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/IntegrationEditorStepAdder.tsx
@@ -83,35 +83,18 @@ export class IntegrationEditorStepAdder extends React.Component<IIntegrationEdit
     this.handleDeleteConfirm = this.handleDeleteConfirm.bind(this);
   }
 
-  public handleDeleteConfirm(
-    integration?: Integration,
-    flowId?: string,
-    position?: number,
-  ) {
-    console.log('position: ' + position);
-    console.log('flowId: ' + flowId);
-    console.log('integration: ' + JSON.stringify(integration));
-
+  public handleDeleteConfirm() {
     if (this.state.showDeleteDialog) {
       this.closeDeleteDialog();
-      /**
-       removeStepFromFlow(
-       integration!,
-       flowId!,
-       position!
-       );
-       // this.onDeleteItem(name);
-       **/
     }
   }
 
-  public onDelete(idx: any) {
-    console.log('idx: ' + idx);
-    this.setState({ showDeleteDialog: true });
+  public onDelete(idx: any): void {
+    this.setState({ position: idx, showDeleteDialog: true });
   }
 
-  public closeDeleteDialog() {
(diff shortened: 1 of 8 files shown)


app/ui-react/syndesis/src/modules/integrations/components/IntegrationEditorStepAdder.tsx
```diff
@@ -45,74 +60,145 @@ export interface IIntegrationEditorStepAdderProps {
...
+                      'integrations:editor:confirmDeleteStepDialogBody'
+                    )}
+                    i18nTitle={t(
+                      'integrations:editor:confirmDeleteStepDialogTitle'
+                    )}
+                    showDialog={this.state.showDeleteDialog}
+                    onCancel={this.closeDeleteDialog}
+                    onConfirm={() => {
+                      this.handleDeleteConfirm();
+                      removeStepFromFlow(
```
gashcrumb: yeah, this here function returns a copy of the integration object with that step removed from the indicated flow.  This is where I guess we need to figure out how to get the page to see the return value of this function.

kahboom: Just to confirm then, what you are saying is that if I set that function as a local var, it should return one less step than the original integration, correct? But I do this:
```
onConfirm={() => {
                      console.log('oldInt: ' + JSON.stringify(this.props.integration));
                      this.handleDeleteConfirm();
                      const newInt = removeStepFromFlow(
                        this.props.integration!,
                        this.props.flowId!,
                        this.state.position!
                      );
                      console.log('newInt: ' + JSON.stringify(newInt));
                    }}
```
And the old integration's `steps` array count is the same as the new one. I would expect one less step for the returned copy of the integration object. I tested that the data passed is correct, so I want to make sure the function is working as expected.

kahboom: Got it, I tested both `const steps = [...flow!.steps!];` (current implementation) and `const steps = getSteps(integration, flowId);` (my changes) for the steps, and it seems to work with my changes. Moving to the next step now (updating the page).

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1532/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #314, opened 2019-05-18
fix: api provider detection and atlasmap url composition

riccardo-forina:
Fix the API provider detection logic.
Fix the URL composition for Atlasmap, checking if we are dealing with absolute or relative paths and act accordingly.

CC @igarashitm one thing that's not clear to me if the `baseMappingServiceUrl` property in the configuration plays a role or not with the other URLs - `baseJavaInspectionServiceUrl`, `baseXMLInspectionServiceUrl`, `baseJSONInspectionServiceUrl`. Right now the logic is that if `baseMappingServiceUrl` is not contained in the specialized URL, we prepend it to it. Logic [is here](https://github.com/syndesisio/syndesis-react/compare/master...riccardo-forina:bugfixes?expand=1#diff-b9d98e0dd7b4c930d63d5c031dff8d6aR84)

Does it sound correct to you?


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1576/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #318, opened 2019-05-19
feat: add navigating away confirmation when in the integration editor

riccardo-forina:
This adds a reusable component `WithLeaveConfirmation` that can be used to protect a page/section from navigating away. 

In the integration editor, it's used by setting the `shouldDisplayDialog` property to a function that checks if the following URL is outside the "mountpoint" of the editor.

## Demo

route-guard

### DIFF
--- a/app/ui-react/syndesis/package.json
+++ b/app/ui-react/syndesis/package.json
@@ -9,6 +9,7 @@
     ]
   },
   "dependencies": {
+    "@allpro/react-router-pause": "^1.1.3",
     "@craco/craco": "^5.0.2",
     "@syndesis/api": "*",
     "@syndesis/apicurio-adapter": "*",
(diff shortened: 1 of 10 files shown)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1600/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

gashcrumb:
Awesome!  I think from the add step page we could skip the confirmation if the integration hasn't been modified, but we could save that for a future PR too, like one for fixing an issue I need to raise still about showing some kind of dirty/modified state with the save/publish buttons :smile:

riccardo-forina:
Yeah this was more a quick win than anything, since the way it’s wired child pages can’t really control the logic.

PR #330, opened 2019-05-21
308 conditional delete integration editor

### LINKED ISSUE #308: Add conditions to delete step in integration editor
relates to #259 

## This is a...

<pre><code>
[ ] Feature request
[ x ] Bug report  
</code></pre>

## Description
Currently, a confirmation dialog is displayed and the step selected is deleted as expected, but these conditions should be obeyed and are not:

- [x] For steps at the start and end of the flow the step should be removed but the user should then be brought to the step select page for that position and be required to set a step/connection and configure it.

- [x] For steps in the middle positions after confirmation the step is simply removed from the step array.

- [ ] Special steps such as the API provider start/finish connections and the conditional flow start/finish connections cannot be removed, so the trash icon should be either hidden or disabled. The angular app hides them.

EDIT: The third item is partially complete. The check and delete works fine but the redirect is not. WIP PR open.

kahboom:
Changes:
- Check and hide ability to delete if step is API provider or CBR
- Move most logic to `AddStepPage` for ease of use with history

Outstanding:
- Check for first and last step of integration flow, on confirmation of delete redirect the user to that position step select page. The check works fine, but the redirect does not. See comments below.

fix #308

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/AddStepPage.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/AddStepPage.tsx
@@ -74,19 +74,19 @@ export class AddStepPage extends React.Component<IAddStepPageProps> {
                             { integration }
                           )
                         }
-                        addStepHref={position =>
-                          this.props.getEditAddStepHref(
-                            position,
-                            { flowId },
-                            { integration }
-                          )
-                        }
                         configureStepHref={(position: number, step: Step) =>
                           getStepHref(
                             step,
                             { flowId, position: `${position}` },
                             { integration },
                             this.props
+                          )
+                        }
+                        editAddStepHref={position =>
+                          this.props.getEditAddStepHref(
+                            position,
+                            { flowId },
+                            { integration }
                           )
                         }
                         flowId={flowId}
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/AddStepPage.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/AddStepPage.tsx
@@ -32,7 +32,7 @@ export interface IAddStepPageProps extends IGetStepHrefs {
     p: IBaseRouteParams,
     s: IBaseRouteState
   ) => H.LocationDescriptor;
-  getEditStepHref?: (
+  getEditStepHref: (
     position: number,
     p: IBaseRouteParams,
     s: IBaseRouteState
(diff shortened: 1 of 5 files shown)


app/ui-react/syndesis/src/modules/integrations/components/editor/AddStepPage.tsx
```diff
@@ -42,72 +63,176 @@ export interface IAddStepPageProps extends IGetStepHrefs {
...
+                           * subsequently redirect the user to the step select
+                           * page for that position.
+                           */
+                          if (
+                            this.state.position ===
+                              getFirstPosition(integration, flowId) ||
+                            this.state.position ===
+                              getLastPosition(integration, flowId)
+                          ) {
+                            history.push(
```
kahboom: @gashcrumb @riccardo-forina - For some reason, when the redirect takes place it results in a blank page and gives the following errors: 
```
TypeError: Cannot read property 'description' of undefined toUIStep 
  87 | // this step is a Connection step
  88 | return {
  89 |   ...step,
> 90 |   description:
     | ^  91 |     (step as ConnectionOverview).description ||
  92 |     step.connection!.description ||
src/modules/integrations/components/editor/utils.ts:90
src/modules/integrations/components/editor/SelectConnectionPage.tsx:70
 70 |   steps: toUIStepCollection(integrationSteps),
```

It worked fine when it was just being done before confirming deletion from within the dialog (so, from line 117 or so). The difference seems to be that I'm using the "state"-defined position.  Any ideas on how I can get around this? Thanks!

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1678/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/integrations/components/editor/AddStepPage.tsx
kahboom: After some experimenting, I think it's not to do with the state-defined position, rather, the step itself. Looking at the Angular app doesn't seem to help a ton. Still trying out a few things.

gashcrumb: So thinking through this more, in the old editor the way I tackled this was to check if the step to be deleted was at the start or at the end of the step array, and if so instead of actually deleting the step, I'd simple overwrite it with an empty step (basically setting it as `{ id: key() }`.  So this may be the same approach we want to take here, as we're not really "adding" a step exactly, instead we're replacing the step with an unconfigured/empty one.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1785/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #332, opened 2019-05-21
feat(logout): allow users to logout

seanforyou23:
This PR adds the ability for users to logout from the user menu.

Screen Shot 2019-05-21 at 3 07 21 PM

It also improves the way we pull in patternfly styles. Previously we had our base styles loaded after the component styles in the DOM, hadn't been able to figure out why before but in working with some others we found that this import `import { UnrecoverableError } from '@syndesis/ui';` coming before we pull in App.css causes the styles to be imported in the wrong order in the DOM. Now with this coming after App.tsx is pulled in, styles are _automatically_ pulled in via the respective react components and we no longer need to import individual patternfly stylesheets as we were doing previously.

I also ditched the approach of using individual base patternfly stylesheets, now that we've made our styling adjustments to the elements that were impacted by introducing PF4 in general, we should be able to safely use the full base stylesheet as provided by react-core.

### DIFF
--- a/app/ui-react/packages/api/src/WithUserHelpers.tsx
+++ b/app/ui-react/packages/api/src/WithUserHelpers.tsx
@@ -26,12 +26,6 @@ export class WithUserHelpersWrapped extends React.Component<
       method: 'GET',
       url: `${process.env.PUBLIC_URL}/logout`,
     });
-
-    // const html = String.fromCharCode.apply(null, new Uint8Array(response.body!));
-    // window.history.pushState(null, null, '/logout');
-    // window.document.open();
-    // window.document.write(html);
-    // window.document.close();
     return await response.body;
   }
 
--- a/app/ui-react/packages/api/src/WithUserHelpers.tsx
+++ b/app/ui-react/packages/api/src/WithUserHelpers.tsx
@@ -24,7 +24,7 @@ export class WithUserHelpersWrapped extends React.Component<
         responseType: 'arraybuffer',
       },
       method: 'GET',
-      url: `/logout`,
+      url: `${process.env.PUBLIC_URL}/logout`,
     });
 
     // const html = String.fromCharCode.apply(null, new Uint8Array(response.body!));
--- a/app/ui-react/packages/api/src/WithUserHelpers.tsx
+++ b/app/ui-react/packages/api/src/WithUserHelpers.tsx
@@ -24,13 +24,15 @@ export class WithUserHelpersWrapped extends React.Component<
         responseType: 'arraybuffer',
       },
       method: 'GET',
-      url: `${this.props.apiUri}/logout`,
+      url: `/logout`,
     });
-    const html = String.fromCharCode.apply(null, new Uint8Array(response.body));
-    window.history.pushState(null, null, '/logout');
-    window.document.open();
-    window.document.write(html);
-    window.document.close();
+
+    // const html = String.fromCharCode.apply(null, new Uint8Array(response.body!));
+    // window.history.pushState(null, null, '/logout');
+    // window.document.open();
+    // window.document.write(html);
+    // window.document.close();
+    return await response.body;
   }
 
   public render() {
--- a/app/ui-react/packages/api/src/WithUserHelpers.tsx
+++ b/app/ui-react/packages/api/src/WithUserHelpers.tsx
@@ -1,0 +1,51 @@
+import * as React from 'react';
+import { ApiContext, IApiContext } from './ApiContext';
+import { callFetch } from './callFetch';
+
+export interface IWithUserHelpersChildrenProps {
+  logout(): Promise<any>;
+}
+export interface IWithUserHelpersProps {
+  children(props: IWithUserHelpersChildrenProps): any;
+}
+
+export class WithUserHelpersWrapped extends React.Component<
+  IWithUserHelpersProps & IApiContext
+> {
+  constructor(props: IWithUserHelpersProps & IApiContext) {
+    super(props);
+    this.logout = this.logout.bind(this);
+  }
+
+  public async logout() {
+    const response = await callFetch({
+      headers: {
+        ...this.props.headers,
+        responseType: 'arraybuffer',
+      },
+      method: 'GET',
+      url: `${this.props.apiUri}/logout`,
+    });
+    const html = String.fromCharCode.apply(null, new Uint8Array(response.body));
+    window.history.pushState(null, null, '/logout');
+    window.document.open();
+    window.document.write(html);
+    window.document.close();
+  }
+
+  public render() {
+    return this.props.children({
+      logout: this.logout,
+    });
+  }
+}
+
+export const WithUserHelpers: React.FunctionComponent<
+  IWithUserHelpersProps
+> = props => {
+  return (
+    <ApiContext.Consumer>
+      {apiContext => <WithUserHelpersWrapped {...props} {...apiContext} />}
+    </ApiContext.Consumer>
+  );
+};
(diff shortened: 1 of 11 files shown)


app/ui-react/packages/api/src/WithUserHelpers.tsx
```diff
@@ -0,0 +1,51 @@
...
+        responseType: 'arraybuffer',
+      },
+      method: 'GET',
+      url: `${this.props.apiUri}/logout`,
+    });
+    // console.log(response);
+    if (!response.ok) {
+      throw new Error(response.statusText);
+    }
+    return await response.body;
```
gashcrumb: So I think instead of doing `return await response.body` we're gonna want to do something like:

```
        const html = String.fromCharCode.apply(null, new Uint8Array(response.body));
        window.history.pushState(null, null, '/logout');
        window.document.open();
        window.document.write(html);
        window.document.close();
        return;
```

I think that should do the trick?

seanforyou23: Yea I've been trying to make this work but getting this type error. Any ideas?

Screen Shot 2019-05-22 at 7 39 04 AM

seanforyou23: And the second one there for pushState gives the error `Argument of type 'null' is not assignable to parameter of type 'string'.` I just pushed the latest with what I have, will continue looking into these errors though.

app/ui-react/packages/api/src/WithUserHelpers.tsx
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  public async logout() {
+    const response = await callFetch({
+      headers: {
+        ...this.props.headers,
+        responseType: 'arraybuffer',
+      },
+      method: 'GET',
+      url: `${this.props.apiUri}/logout`,
```
gashcrumb: This should just be `/logout`, no `${this.props.apiUri}` needed.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1703/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

seanforyou23:
Interesting, I'm getting a 200 response for logout locally, but in the PR build I get a 404. In either case, I'm not redirected to a logout page, and if I refresh the page it doesn't look like I actually logged out anyway. Getting close though. Maybe @zregvart has an idea of what the missing piece may be?

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1705/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1723/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1730/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/packages/api/src/WithUserHelpers.tsx
```diff
@@ -0,0 +1,53 @@
...
+  constructor(props: IWithUserHelpersProps & IApiContext) {
+    super(props);
+    this.logout = this.logout.bind(this);
+  }
+
+  public async logout() {
+    const response = await callFetch({
+      headers: {
+        ...this.props.headers,
+        responseType: 'arraybuffer',
```
zregvart:
```suggestion
        'SYNDESIS-XSRF-TOKEN': 'awesome',
        responseType: 'arraybuffer',
```

zregvart:
> Interesting, I'm getting a 200 response for logout locally, but in the PR build I get a 404. In either case, I'm not redirected to a logout page, and if I refresh the page it doesn't look like I actually logged out anyway. Getting close though. Maybe @zregvart has an idea of what the missing piece may be?

I'm guessing that locally you run against staging and on the PR build against stubs?

Thing to watch out for is the `_oauth_proxy` cookie, when you hit logout (if `SYNDESIS-XSRF-TOKEN: awesome` HTTP header is present) it should get cleared when `/logout` endpoint is invoked. If it doesn't get cleared (at that point) then we have a bug. Anyhows see the suggestion on the PR.

Other thing that we need to be aware of: because we have single sign on using OAuth, and we're not doing single logout on the staging cluster against GitHub it can appear as if you didn't logout, but what actually is happening is that you get logged out can get logged in right away (happens really fast since you already approved oauth-proxy's OAuth use).

app/ui-react/packages/api/src/WithUserHelpers.tsx
riccardo-forina: That's not needed, the right headers come from [here](https://github.com/syndesisio/syndesis-react/pull/332/files/19ef60977804468b9d366ed71d52c208000946b6#diff-abcbbbbc168c9e47faf84ef500912928R49)

app/ui-react/packages/api/src/WithUserHelpers.tsx
```diff
@@ -0,0 +1,53 @@
...
+      },
+      method: 'GET',
+      url: `/logout`,
```
riccardo-forina: Ok this is tricky. The logout page, while being a route, it's not _an app_ route, so we should handle it like an asset. I think this could help:

```tsx
url: `${process.env.PUBLIC_URL}/logout`, 
```

The thing is, on CI assets are not in the root folder, so we have to remember to always append the `PUBLIC_URL` bit to the url.

app/ui-react/syndesis/src/app/UI.tsx
```diff
@@ -200,18 +199,24 @@ export const UI: React.FunctionComponent<IAppUIProps> = ({ routes }) => {
...
-                              </AppTopMenu>
+                              <WithUserHelpers>
+                                {({ logout }) => {
+                                  return (
+                                    <AppTopMenu
+                                      username={user.username || 'developer'}
+                                      onSelectLogout={() => {
+                                        logout().then(() => {
+                                          console.log(
+                                            'do we need to manually redirect here?'
```
gashcrumb: I believe we need not do anything except resolve the promise really, the logout operation will make the browser navigate away from the app.

app/ui-react/packages/api/src/WithUserHelpers.tsx
seanforyou23: Yep, I do see `SYNDESIS-XSRF-TOKEN` being provided as part of the request headers 👍

seanforyou23:
Thx for the feedback everyone. So in the _response_ headers I see

`set-cookie: _oauth_proxy=deleted; Expires=Thu, 01-Jan-1970 00:00:01 GMT; Domain=.syndesis-staging.b6ff.rh-idev.openshiftapps.com; HttpOnly; Secure` - which I think looks fine.

But in the _request_ headers I see only `Cookie: io=qfXo2D4soeosanqsAAAD` which looks like it's missing some things (comparing against staging w/ angular app) I think that's my hang up.

Is the [proxy definition](https://github.com/syndesisio/syndesis-react/pull/332/files#diff-455d3a37798181348f180667b8f40514R26) ok? I used the same format as the `/api/v1` endpoint and swapped the path to /logout. Should I be doing something different with the cookie here?

gashcrumb:
> Is the proxy definition ok? I used the same format as the /api/v1 endpoint and swapped the path to /logout. Should I be doing something different with the cookie here?

I think it's fine, and no the proxy should only just be proxying.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1788/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #359, opened 2019-05-27
fix: revert bogus changes, fixes #348

### LINKED ISSUE #348: describe-data routing appears to lose the correct step index
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
Noticed this when testing out something else, so to reproduce create an integration using an AMQ connection.  Use "Subscribe for messages" and you'll be at this config page:

image

'Next' takes you to the "describe data page" but look at what happens in the side-bar:

image

Then when I click next again:

image

so I think now I'm filling in the end step, but I picked something:

image

And now we're here:

image

the last step there doesn't seem to edit properly either, in that if I change the action to an appropriate one I don't wind up at the describe data page after the configuration page.

riccardo-forina:
Fixes #348

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/endpoint/ConfigureActionPage.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/endpoint/ConfigureActionPage.tsx
@@ -144,8 +144,9 @@ export class ConfigureActionPage extends React.Component<
                         },
                         {
                           connection,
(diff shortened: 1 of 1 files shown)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/1853/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #387, opened 2019-05-31
TTOOLS-646 added save button enablement logic and exposed validation …

blafond:
…response error message from Teiid parsing failure


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2049/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

blafond:
> It looks like you can eliminate the 'ddlChanged' state and just use 'needsValidation' when the ddl changes?
> Only other thing is spelling error - use 'Validation' instead of 'Valiation'

I actually tried to use just one of those, but the initial state requires both if we want SAVE disabled when the page comes up with valid DDL

mdrillin:
Ok, works for me...

PR #391, opened 2019-06-03
fix: show the right steps in the editor

riccardo-forina:
We were missing a call to the API that fetches the updated descriptor for steps, now we have that.
This fixes - together with some other changes - the steps we display in the editor.

Steps to reproduce:
1. new integration: select * from todo > log
2. add split step in between
3. add a step after the split, basic filter should now show up (it wasn't showing up before)

Also actions like adding/updating/removing steps are all now async, so we show the right spinners where we should.


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2084/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #374, opened 2019-05-30
The connections list page's disabled 'delete' menu item no longer triggers a delete confirmation

elvisisking:
- See #208
- `ConnectionCard` now uses PF4 Dropdown
- Fixes a regression bug in `ConnectionDetailsForm` that kept the cancel and save buttons always visible
- Added property to `PfDropdownItem` that indicates if item should be disabled
- Added a `ConnectionDetailsForm` story

PR #409, opened 2019-06-04
Tech Preview feature not working for Integration Editor

kahboom:
Created another version of `ConnectionsWithToolbar` per @riccardo-forina 's suggestion.

Screenshot 2019-06-04 16 06 10

fix #374

### DIFF
--- a/app/ui-react/packages/api/src/helpers/connectionFunctions.ts
+++ b/app/ui-react/packages/api/src/helpers/connectionFunctions.ts
@@ -8,6 +8,7 @@ import {
   ConnectionOverview,
   Connector,
 } from '@syndesis/models';
+import { getMetadataValue } from './integrationFunctions';
 
 export function getActionsWithFrom(actions: Action[] = []) {
   return actions.filter(a => a.pattern === 'From');
@@ -91,6 +92,19 @@ export function isConfigurationRequired(
   board: ConnectionBulletinBoard
 ): boolean {
   return (board!.notices || board!.warnings || board!.errors)! > 0;
+}
+
+/**
+ * Checks whether or not the provided object is a technical preview.
+ * Accepts a Connector.
+ * Returns a boolean for whether or not the metadata `tech-preview` key
+ * returns a string value of 'true'
+ * @param connector
+ */
+export function isTechPreview(connector: Connector): boolean {
+  return (
+    getMetadataValue<string>('tech-preview', connector.metadata) === 'true'
+  );
 }
 
 /**
--- a/app/ui-react/packages/api/src/helpers/connectionFunctions.ts
+++ b/app/ui-react/packages/api/src/helpers/connectionFunctions.ts
@@ -91,7 +91,7 @@ export function getActionStepDefinition(
 export function isConfigurationRequired(
   board: ConnectionBulletinBoard
 ): boolean {
-  return (board!.notices || board!.warnings || board!.errors)! > 0;
+  return (board.notices || board.warnings || board.errors)! > 0;
 }
 
 /**
--- a/app/ui-react/packages/api/src/helpers/connectionFunctions.ts
+++ b/app/ui-react/packages/api/src/helpers/connectionFunctions.ts
@@ -83,6 +83,10 @@ export function getActionStepDefinition(
   return step.properties;
 }
 
+/**
+ * Checks whether the ConnectionBulletinBoard provided
+ * @param board
+ */
 export function isConfigurationRequired(
   board: ConnectionBulletinBoard
 ): boolean {
--- a/app/ui-react/packages/api/src/helpers/connectionFunctions.ts
+++ b/app/ui-react/packages/api/src/helpers/connectionFunctions.ts
@@ -5,8 +5,8 @@ import {
   ConfigurationProperty,
   Connection,
   ConnectionBulletinBoard,
-  ConnectionOverview,
-  Connector,
+  IConnectionOverview,
+  IConnector,
 } from '@syndesis/models';
 import { getMetadataValue } from './integrationFunctions';
 
@@ -27,15 +27,15 @@ export function getConnectionMetadataValue(
 }
 
 export function getConnectionConnector(
-  connection: ConnectionOverview
-): Connector {
+  connection: IConnectionOverview
+): IConnector {
   if (!connection.connector) {
     throw Error(`FATAL: Connection ${connection.id} doesn't have a connector`);
   }
   return connection.connector;
 }
 
-export function getConnectorActions(connector: Connector): Action[] {
+export function getConnectorActions(connector: IConnector): Action[] {
   return connector.actions;
 }
 
@@ -101,7 +101,7 @@ export function isConfigurationRequired(
  * returns a string value of 'true'
  * @param connector
  */
-export function isTechPreview(connector: Connector): boolean {
+export function isTechPreview(connector: IConnector): boolean {
   return (
     getMetadataValue<string>('tech-preview', connector.metadata) === 'true'
   );
--- a/app/ui-react/packages/api/src/helpers/connectionFunctions.ts
+++ b/app/ui-react/packages/api/src/helpers/connectionFunctions.ts
@@ -4,6 +4,7 @@ import {
   ActionDescriptorStep,
   ConfigurationProperty,
   Connection,
+  ConnectionBulletinBoard,
   ConnectionOverview,
   Connector,
 } from '@syndesis/models';
@@ -80,6 +81,12 @@ export function getActionStepDefinition(
     throw Error(`FATAL: Step ${step} does not have valid properties`);
   }
   return step.properties;
+}
+
+export function isConfigurationRequired(
+  board: ConnectionBulletinBoard
+): boolean {
+  return (board!.notices || board!.warnings || board!.errors)! > 0;
 }
 
 /**
--- a/app/ui-react/packages/api/src/helpers/connectionFunctions.ts
+++ b/app/ui-react/packages/api/src/helpers/connectionFunctions.ts
@@ -88,9 +88,7 @@ export function getActionStepDefinition(
  * Checks whether the ConnectionBulletinBoard provided
  * @param board
  */
-export function isConfigurationRequired(
-  board: ConnectionBulletinBoard
-): boolean {
+export function isConfigRequired(board: ConnectionBulletinBoard): boolean {
   return (board.notices || board.warnings || board.errors)! > 0;
 }
 
--- a/app/ui-react/packages/api/src/useConnection.tsx
+++ b/app/ui-react/packages/api/src/useConnection.tsx
@@ -1,5 +1,10 @@
 import { ConnectionOverview } from '@syndesis/models';
-import { getActionsWithFrom, getActionsWithTo, isDerived } from './helpers';
+import {
+  getActionsWithFrom,
+  getActionsWithTo,
+  isConfigurationRequired,
+  isDerived,
+} from './helpers';
 import { useApiResource } from './useApiResource';
 
 export const transformConnectionResponse = (connection: ConnectionOverview) => {
@@ -11,6 +16,7 @@ export const transformConnectionResponse = (connection: ConnectionOverview) => {
     actionsWithTo: getActionsWithTo(
       connection.connector ? connection.connector.actions : []
     ),
+    configRequired: isConfigurationRequired(connection),
     derived: isDerived(connection),
   };
 };
--- a/app/ui-react/packages/api/src/useConnection.tsx
+++ b/app/ui-react/packages/api/src/useConnection.tsx
@@ -19,7 +19,7 @@ export const transformConnectionResponse = (connection: ConnectionOverview) => {
     ),
     configRequired: isConfigurationRequired(connection),
     derived: isDerived(connection),
-    isTechPreview: isTechPreview(connection.connector),
+    isTechPreview: isTechPreview(connection.connector!),
   };
 };
 
--- a/app/ui-react/packages/api/src/useConnection.tsx
+++ b/app/ui-react/packages/api/src/useConnection.tsx
@@ -4,6 +4,7 @@ import {
   getActionsWithTo,
   isConfigurationRequired,
   isDerived,
+  isTechPreview,
 } from './helpers';
 import { useApiResource } from './useApiResource';
 
@@ -18,6 +19,7 @@ export const transformConnectionResponse = (connection: ConnectionOverview) => {
     ),
     configRequired: isConfigurationRequired(connection),
     derived: isDerived(connection),
+    isTechPreview: isTechPreview(connection.connector),
   };
 };
 
--- a/app/ui-react/packages/api/src/useConnection.tsx
+++ b/app/ui-react/packages/api/src/useConnection.tsx
@@ -1,4 +1,4 @@
-import { ConnectionOverview } from '@syndesis/models';
+import { IConnectionOverview } from '@syndesis/models';
 import {
   getActionsWithFrom,
   getActionsWithTo,
@@ -8,7 +8,9 @@ import {
 } from './helpers';
 import { useApiResource } from './useApiResource';
 
-export const transformConnectionResponse = (connection: ConnectionOverview) => {
+export const transformConnectionResponse = (
+  connection: IConnectionOverview
+) => {
   return {
     ...connection,
     actionsWithFrom: getActionsWithFrom(
@@ -25,9 +27,9 @@ export const transformConnectionResponse = (connection: ConnectionOverview) => {
 
 export const useConnection = (
   connectionId: string,
-  initialValue?: ConnectionOverview
+  initialValue?: IConnectionOverview
 ) => {
-  return useApiResource<ConnectionOverview>({
+  return useApiResource<IConnectionOverview>({
     defaultValue: {
       name: '',
     },
--- a/app/ui-react/packages/api/src/useConnection.tsx
+++ b/app/ui-react/packages/api/src/useConnection.tsx
@@ -2,7 +2,7 @@ import { IConnectionOverview } from '@syndesis/models';
 import {
   getActionsWithFrom,
   getActionsWithTo,
-  isConfigurationRequired,
+  isConfigRequired,
   isDerived,
 } from './helpers';
 import { useApiResource } from './useApiResource';
@@ -18,9 +18,9 @@ export const transformConnectionResponse = (
     ...connection,
     actionsWithFrom: getActionsWithFrom(connector ? connector.actions : []),
     actionsWithTo: getActionsWithTo(connector ? connector.actions : []),
-    configRequired: isConfigurationRequired(connection),
     connector,
     derived: isDerived(connection),
+    isConfigRequired: isConfigRequired(connection),
     isTechPreview: connector ? connector.isTechPreview : false,
   };
 };
@@ -31,7 +31,7 @@ export const useConnection = (
 ) => {
   return useApiResource<IConnectionOverview>({
     defaultValue: {
-      configRequired: false,
+      isConfigRequired: false,
       isTechPreview: false,
       name: '',
     },
--- a/app/ui-react/packages/api/src/useConnection.tsx
+++ b/app/ui-react/packages/api/src/useConnection.tsx
@@ -4,24 +4,24 @@ import {
   getActionsWithTo,
   isConfigurationRequired,
   isDerived,
-  isTechPreview,
 } from './helpers';
 import { useApiResource } from './useApiResource';
+import { transformConnectorResponse } from './useConnector';
 
 export const transformConnectionResponse = (
   connection: IConnectionOverview
 ) => {
+  const connector = connection.connector
+    ? transformConnectorResponse(connection.connector)
+    : undefined;
   return {
     ...connection,
-    actionsWithFrom: getActionsWithFrom(
-      connection.connector ? connection.connector.actions : []
-    ),
-    actionsWithTo: getActionsWithTo(
-      connection.connector ? connection.connector.actions : []
-    ),
+    actionsWithFrom: getActionsWithFrom(connector ? connector.actions : []),
+    actionsWithTo: getActionsWithTo(connector ? connector.actions : []),
     configRequired: isConfigurationRequired(connection),
+    connector,
     derived: isDerived(connection),
-    isTechPreview: isTechPreview(connection.connector!),
+    isTechPreview: connector ? connector.isTechPreview : false,
   };
 };
 
--- a/app/ui-react/packages/api/src/useConnection.tsx
+++ b/app/ui-react/packages/api/src/useConnection.tsx
@@ -31,6 +31,8 @@ export const useConnection = (
 ) => {
   return useApiResource<IConnectionOverview>({
     defaultValue: {
+      configRequired: false,
+      isTechPreview: false,
       name: '',
     },
     initialValue,
--- a/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
+++ b/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
@@ -12,7 +12,7 @@ import * as React from 'react';
 import { Translation } from 'react-i18next';
 import { UIContext } from '../../../app';
 import { ApiError, EntityIcon } from '../../../shared';
-import { figureOutTechPreviewFlag } from '../../integrations/components/editor/utils';
+import { isTechPreview } from '../../integrations/components/editor/utils';
 
 export interface IConnectionsProps {
   error: boolean;
@@ -81,7 +81,7 @@ export class Connections extends React.Component<IConnectionsProps> {
                               const configurationRequired =
                                 c.board && isConfigurationRequired(c.board);
 
-                              const isTechPreview = figureOutTechPreviewFlag(c);
+                              const techPreview = isTechPreview(c);
 
                               return (
                                 <ConnectionsGridCell key={index}>
@@ -129,7 +129,7 @@ export class Connections extends React.Component<IConnectionsProps> {
                                           }
                                         : undefined
                                     }
-                                    techPreview={isTechPreview}
+                                    techPreview={techPreview}
                                     techPreviewPopoverHtml={
                                       <span
                                         dangerouslySetInnerHTML={{
--- a/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
+++ b/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
@@ -81,7 +81,6 @@ export class Connections extends React.Component<IConnectionsProps> {
                                 <ConnectionsGridCell key={index}>
                                   <ConnectionCard
                                     name={c.name}
-                                    configurationRequired={c.configRequired}
                                     description={c.description || ''}
                                     icon={
                                       <EntityIcon
@@ -92,10 +91,12 @@ export class Connections extends React.Component<IConnectionsProps> {
                                     }
                                     href={this.props.getConnectionHref(c)}
                                     i18nCannotDelete={t('cannotDelete')}
-                                    i18nConfigurationRequired={t(
+                                    i18nConfigRequired={t(
                                       'configurationRequired'
                                     )}
                                     i18nTechPreview={t('techPreview')}
+                                    isConfigRequired={c.isConfigRequired}
+                                    isTechPreview={c.isTechPreview}
                                     menuProps={
                                       this.props.includeConnectionMenu
                                         ? {
@@ -121,7 +122,6 @@ export class Connections extends React.Component<IConnectionsProps> {
                                           }
                                         : undefined
                                     }
-                                    techPreview={c.isTechPreview}
                                     techPreviewPopoverHtml={
                                       <span
                                         dangerouslySetInnerHTML={{
--- a/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
+++ b/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
@@ -1,4 +1,4 @@
-import { isConfigurationRequired, WithConnectionHelpers } from '@syndesis/api';
+import { WithConnectionHelpers } from '@syndesis/api';
 import * as H from '@syndesis/history';
 import { ConnectionOverview } from '@syndesis/models';
 import {
@@ -12,7 +12,6 @@ import * as React from 'react';
 import { Translation } from 'react-i18next';
 import { UIContext } from '../../../app';
 import { ApiError, EntityIcon } from '../../../shared';
-import { isTechPreview } from '../../integrations/components/editor/utils';
 
 export interface IConnectionsProps {
   error: boolean;
@@ -78,20 +77,11 @@ export class Connections extends React.Component<IConnectionsProps> {
                                 doDelete(c.id!, c.name); // must have an ID if deleting
                               };
 
-                              const configurationRequired =
-                                c.board && isConfigurationRequired(c.board);
-
-                              const techPreview = c.connector
-                                ? isTechPreview(c.connector)
-                                : false;
-
                               return (
                                 <ConnectionsGridCell key={index}>
                                   <ConnectionCard
                                     name={c.name}
-                                    configurationRequired={
-                                      configurationRequired
-                                    }
+                                    configurationRequired={c.configRequired}
                                     description={c.description || ''}
                                     icon={
                                       <EntityIcon
@@ -131,7 +121,7 @@ export class Connections extends React.Component<IConnectionsProps> {
                                           }
                                         : undefined
                                     }
-                                    techPreview={techPreview}
+                                    techPreview={c.isTechPreview}
                                     techPreviewPopoverHtml={
                                       <span
                                         dangerouslySetInnerHTML={{
--- a/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
+++ b/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
@@ -1,6 +1,6 @@
 import { WithConnectionHelpers } from '@syndesis/api';
 import * as H from '@syndesis/history';
-import { ConnectionBulletinBoard, ConnectionOverview } from '@syndesis/models';
+import { ConnectionOverview, IConnectionOverview } from '@syndesis/models';
 import {
   ConnectionCard,
   ConnectionsGrid,
@@ -12,12 +12,6 @@ import * as React from 'react';
 import { Translation } from 'react-i18next';
 import { UIContext } from '../../../app';
 import { ApiError, EntityIcon } from '../../../shared';
-
-export interface IConnectionOverview extends ConnectionOverview {
-  board?: ConnectionBulletinBoard;
-  configRequired?: boolean;
-  isTechPreview?: boolean;
-}
 
 export interface IConnectionsProps {
   error: boolean;
--- a/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
+++ b/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
@@ -81,7 +81,9 @@ export class Connections extends React.Component<IConnectionsProps> {
                               const configurationRequired =
                                 c.board && isConfigurationRequired(c.board);
 
-                              const techPreview = isTechPreview(c);
+                              const techPreview = c.connector
+                                ? isTechPreview(c.connector)
+                                : false;
 
                               return (
                                 <ConnectionsGridCell key={index}>
--- a/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
+++ b/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
@@ -1,6 +1,6 @@
 import { WithConnectionHelpers } from '@syndesis/api';
 import * as H from '@syndesis/history';
-import { ConnectionOverview, IConnectionOverview } from '@syndesis/models';
+import { IConnectionOverview } from '@syndesis/models';
 import {
   ConnectionCard,
   ConnectionsGrid,
@@ -19,8 +19,8 @@ export interface IConnectionsProps {
   loading: boolean;
   connections: IConnectionOverview[];
 
-  getConnectionHref(connection: ConnectionOverview): H.LocationDescriptor;
-  getConnectionEditHref?(connection: ConnectionOverview): H.LocationDescriptor;
+  getConnectionHref(connection: IConnectionOverview): H.LocationDescriptor;
+  getConnectionEditHref?(connection: IConnectionOverview): H.LocationDescriptor;
 }
 
 export class Connections extends React.Component<IConnectionsProps> {
--- a/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
+++ b/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
@@ -1,6 +1,6 @@
 import { WithConnectionHelpers } from '@syndesis/api';
 import * as H from '@syndesis/history';
-import { ConnectionOverview } from '@syndesis/models';
+import { ConnectionBulletinBoard, ConnectionOverview } from '@syndesis/models';
 import {
   ConnectionCard,
   ConnectionsGrid,
@@ -13,11 +13,17 @@ import { Translation } from 'react-i18next';
 import { UIContext } from '../../../app';
 import { ApiError, EntityIcon } from '../../../shared';
 
+export interface IConnectionOverview extends ConnectionOverview {
+  board?: ConnectionBulletinBoard;
+  configRequired?: boolean;
+  isTechPreview?: boolean;
+}
+
 export interface IConnectionsProps {
   error: boolean;
   includeConnectionMenu: boolean;
   loading: boolean;
-  connections: ConnectionOverview[];
+  connections: IConnectionOverview[];
 
   getConnectionHref(connection: ConnectionOverview): H.LocationDescriptor;
   getConnectionEditHref?(connection: ConnectionOverview): H.LocationDescriptor;
--- a/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
+++ b/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
@@ -1,4 +1,4 @@
-import { WithConnectionHelpers } from '@syndesis/api';
+import { isConfigurationRequired, WithConnectionHelpers } from '@syndesis/api';
 import * as H from '@syndesis/history';
 import { ConnectionOverview } from '@syndesis/models';
 import {
@@ -12,6 +12,7 @@ import * as React from 'react';
 import { Translation } from 'react-i18next';
 import { UIContext } from '../../../app';
 import { ApiError, EntityIcon } from '../../../shared';
+import { figureOutTechPreviewFlag } from '../../integrations/components/editor/utils';
 
 export interface IConnectionsProps {
   error: boolean;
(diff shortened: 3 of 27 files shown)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2146/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/connections/components/ConnectionsWithToolbarNew.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+const sortTypes: ISortType[] = [sortByName];
+
+export interface IConnectionsWithToolbarNewProps
+  extends IConnectionsProps,
+    Pick<IConnectionsListViewProps, 'createConnectionButtonStyle'> {
+  children?: any;
+}
+
+export class ConnectionsWithToolbarNew extends React.Component<
```
gashcrumb: U need a better name.  I wonder, maybe this should live in the `integrations` module and maybe be called like I dunno, `AllStepsWithToolbar` or something?  That way it's a bit more obvious that this guy isn't just for "connections" in the syndesis sense.  Ooh ooh, maybe `UIStepsWithToolbar`

gashcrumb (changes requested):
The name, it has to change :smile:

app/ui-react/syndesis/src/modules/connections/components/ConnectionsWithToolbarNew.tsx
gashcrumb: Works great BTW!

kahboom: Yeah I was struggling a bit with the name, as you can see. 😂 I think `UIStepsWithToolbar` sounds good. 👍

app/ui-react/syndesis/src/modules/connections/components/ConnectionsWithToolbarNew.tsx
```diff
@@ -0,0 +1,126 @@
...
+} from '@syndesis/ui';
+import { WithListViewToolbarHelpers } from '@syndesis/utils';
+import * as React from 'react';
+import { Translation } from 'react-i18next';
+import i18n from '../../../i18n';
+import resolvers from '../resolvers';
+import { Connections, IConnectionsProps } from './Connections';
+
+function getFilteredAndSortedConnections(
+  connections: Connection[],
```
riccardo-forina: we are not dealing with `Connection` objects here, but with `IUIIntegrationStep` which is a "meta object" that can hold `Connection`s, `StepKind`s and `Extension`s

app/ui-react/syndesis/src/modules/connections/components/ConnectionsWithToolbarNew.tsx
```diff
@@ -0,0 +1,126 @@
...
+                  sortTypes={sortTypes}
+                  resultsCount={filteredAndSortedConnections.length}
+                  {...helpers}
+                  i18nLinkCreateConnection={t('shared:linkCreateConnection')}
+                  i18nResultsCount={t('shared:resultsCount', {
+                    count: filteredAndSortedConnections.length,
+                  })}
+                >
+                  {this.props.children}
+                  <Connections
```
riccardo-forina: Because of the above, we can't reuse the `Connections` component.

riccardo-forina (changes requested):
I suggest `EditorStep` for the name.

kahboom:
Alright, here are the changes I've made, based on feedback, that seem to work (albeit some are hacky):
- Rename to `EditorStepsWithToolbar`
- Move to `integrations` module for reuse
- Duplicate of `Connections` component called `EditorSteps` that uses the `IUIIntegrationStep` interface and `EditorStepsWithToolbar`
- Adjusted the typings to comply (please note the use of `as Blah[]` a few times, not ideal but it works for now)
- Rename `connections` to `steps` to reflect data more accurately

And still using the `.map()` part for the `filteredAndSortedConnections`, as we needed it.

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
```diff
@@ -0,0 +1,166 @@
...
+                                            c
+                                          )
+                                    }
+                                    href={this.props.getConnectionHref(c)}
+                                    i18nCannotDelete={t('cannotDelete')}
+                                    i18nConfigurationRequired={t(
+                                      'configurationRequired'
+                                    )}
+                                    i18nTechPreview={t('techPreview')}
+                                    menuProps={
```
riccardo-forina: I don’t think we have to show a menu at all here

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorStepsWithToolbar.tsx
```diff
@@ -0,0 +1,126 @@
...
+  steps: IUIIntegrationStep[],
+  activeFilters: IActiveFilter[],
+  currentSortType: ISortType,
+  isSortAscending: boolean
+) {
+  let filteredAndSortedConnections = steps;
+  activeFilters.forEach((filter: IActiveFilter) => {
+    const valueToLower = filter.value.toLowerCase();
+    filteredAndSortedConnections = filteredAndSortedConnections.filter(
+      (c: Connection) => c.name.toLowerCase().includes(valueToLower)
```
riccardo-forina: This cast doesn’t feel right. Name should be a common property

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorStepsWithToolbar.tsx
```diff
@@ -0,0 +1,126 @@
...
+            {helpers => {
+              const connectionString = 'connection';
+
+              const filteredAndSortedConnections = getFilteredAndSortedConnections(
+                this.props.connections as IUIIntegrationStep[],
+                helpers.activeFilters,
+                helpers.currentSortType,
+                helpers.isSortAscending
+              ).map(connection => {
+                return connection[connectionString];
```
riccardo-forina: Not all items have a connection property I think? I’ll have to check this tomorrow against a BE but I think we might be ending up not showing all the steps

riccardo-forina (commented):
Just a quick check, I’ll test this more thoroughly tomorrow

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorStepsWithToolbar.tsx
gashcrumb: Yeah, I'd suggest continuing to rename these variables for this file, in this case it's `filteredAndSortedEditorSteps` not connections.  and then not `c`, maybe `editorStep` or `es`.

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorStepsWithToolbar.tsx
gashcrumb: Yeah, only `endpoint` kinds would have a `connection` property.  But we could be misled here if  it's just because `connection` needs to be renamed to something like `editorStep`.

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
```diff
@@ -0,0 +1,166 @@
...
+export class EditorSteps extends React.Component<IEditorStepsProps> {
+  public render() {
+    return (
+      <Translation ns={['connections', 'shared']}>
+        {t => (
+          <UIContext.Consumer>
+            {({ pushNotification }) => {
+              return (
+                <WithConnectionHelpers>
+                  {({ deleteConnection }) => {
```
gashcrumb: Delete connection isn't needed here at all, this page doesn't require any management functions.

gashcrumb: That means you can get rid of the notifications, `WithConnectionHelpers` etc

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2164/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorStepsWithToolbar.tsx
kahboom: Renaming to `editorStep[connectionString]`. I can do a check for `uiStepKind === 'editor-step'`, but at the moment only one of all the listed items isn't `endpoint`, and that's `api-provider`. Also, from what I'm seeing, even `api-provider` has a `connection` property. This is how the list appears for me prior to `.map()`: https://gist.github.com/kahboom/d58532286c0f94da46ef17e0c477ba73

In the raw version, grep `api-provider` as an example.

riccardo-forina: `editor-step` [isn't a valid value](https://github.com/syndesisio/syndesis-react/blob/4cbbdfbe57f0e08ea133138551a7752dccacf737/app/ui-react/syndesis/src/modules/integrations/components/editor/interfaces.ts#L34) for uiStepKind.

Again, we are not dealing with connections, but with a mixed bag of different things: connections, stepkinds and extensions. All those types [have been normalized as `IUIStep` objects](https://github.com/syndesisio/syndesis-react/blob/e3b8fc2d0ae9454ce7edf633f5f11afce6994abf/app/ui-react/syndesis/src/modules/integrations/components/editor/SelectConnectionPage.tsx#L102-L106). You'll have a valid `title` - to be used for the sorting an filtering - and for some objects a valid `connector` object. All will have some `metadata` set.
All this info needs to be properly handled to show - or not - the tech preview/configuration required badges.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2188/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2201/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorStepsWithToolbar.tsx
kahboom: I'm aware that we are not dealing with connections, but what I'm getting for the data is an array of objects each with a parent `connection` property.
```
[
  { connection: { configuredProperties: { .. }, connector: { .. }, metadata: { .. }  } },
  { connection: { .. } },
  { connection: { .. } },
  and so on..
]
```
I can either `.map()` it or similar, or I can write some type of interface like
```
export interface IUIEditorStep extends IUIIntegrationStep {
  connection: IUIStep
}
```
Neither is ideal, but if you have a different approach in mind please share what you are looking for exactly.

The original purpose of the PR was to fix the Tech Preview feature that was previously working, if we want to roll out a new toolbar for connections and are changing the object, that's getting kinda out of the scope of this PR.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2222/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorStepsWithToolbar.tsx
gashcrumb: Can you re-read what @riccardo-forina said?  You can't rely on `connection` being set for this particular collection.  Much like the old angular app tried to re-use the "Connection List" page we actually kinda started in the same path here in the React app, as I recall when this page was originally written I had just created the combined step/connection/extension step in the angular app.  So this component that you're writing has to deal with that same combination.  At the very least, these objects are normalized into an IUIStep, so you should only need to look at the `metadata` value for each thing in the collection, not the `connection` object directly.  Also, use the `getMetadataValue<string>`, helper just in the off chance an item in the collection doesn't have it set.  I haven't taken a close look to see if you're using that here.

riccardo-forina (changes requested):
It doesn't work

doesntwork

riccardo-forina:
Actually, this needs a rebase, the error I posted above is because we have an extension on staging and this PR is based on a version of the app that was bugging out with that.

kahboom:
Screenshot 2019-06-06 17 20 10

No additional or changes to interfaces, no `.map`, no referencing the `Connection` interface from what I can see, just added an additional check and used @gashcrumb 's suggestion of the `getMetadataValue` helper. I did read @riccardo-forina 's suggestion several times. I do understand that this is a not a Connection and instead comprises connections, extensions, and stepKinds, where `IUIStep` normalizes it. This is using that interface. Something happens in `mergeConnectionsSources` where a `connection` property gets added to each object. I understand we can't rely on that, but it's still there, and needs to be iterated over to reach `metadata`, so I've added another check in `EditorSteps.tsx` to check if that property exists. Short of that I'm out of ideas, so let me know what exactly would need to change, if anything.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2240/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

riccardo-forina:
Did you test something different than the first step flow? Try adding a 3rd step, maybe in one of the existing integrations, the collection will be a mixed thing. In the first step, only connections and extensions are allowed

kahboom:
Good catch, I should have checked for `connection`, let me know if this suffices.

Screenshot 2019-06-06 19 31 45

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2244/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorStepsWithToolbar.tsx
```diff
@@ -0,0 +1,119 @@
...
+                    count: filteredAndSortedEditorSteps.length,
+                  })}
+                >
+                  {this.props.children}
+                  <EditorSteps
+                    error={this.props.error}
+                    loading={this.props.loading}
+                    steps={filteredAndSortedEditorSteps}
+                    getConnectionHref={this.props.getConnectionHref}
+                    getConnectionEditHref={this.props.getConnectionEditHref}
```
riccardo-forina: This isn't needed

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { Translation } from 'react-i18next';
+import { ApiError, EntityIcon } from '../../../../shared';
+import { IUIStep } from './interfaces';
+
+export interface IEditorStepsProps {
+  error: boolean;
+  loading: boolean;
+  steps: IUIStep[];
+
+  getConnectionHref(connection: ConnectionOverview): H.LocationDescriptor;
```
riccardo-forina: we are not dealing with just connections, so better not use `connection` when naming things. Also, the object that's passed back should really be a `IUIStep`, shouldn't it?

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+                  let isTechPreview = false;
+
+                  if (s.connection && s.connection!.connector) {
+                    isTechPreview =
+                      getMetadataValue<string>(
+                        'tech-preview',
+                        s!.connection!.connector!.metadata
+                      ) === 'true';
+                  } else if (s.connector!) {
```
riccardo-forina: the `!` isn't needed here

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
```diff
@@ -0,0 +1,99 @@
...
+                    isTechPreview =
+                      getMetadataValue<string>(
+                        'tech-preview',
+                        s.connector!.metadata
```
riccardo-forina: same here, the above makes the object valid

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
```diff
@@ -0,0 +1,99 @@
...
+                      s.board!.warnings ||
+                      s.board!.errors)! > 0;
+
+                  let isTechPreview = false;
+
+                  if (s.connection && s.connection!.connector) {
+                    isTechPreview =
+                      getMetadataValue<string>(
+                        'tech-preview',
+                        s!.connection!.connector!.metadata
```
riccardo-forina: drop the `!`, the above `if` is making those superfluous

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorStepsWithToolbar.tsx
```diff
@@ -0,0 +1,119 @@
...
+  steps: IUIStep[],
+  activeFilters: IActiveFilter[],
+  currentSortType: ISortType,
+  isSortAscending: boolean
+) {
+  let filteredAndSortedEditorSteps = steps;
+  activeFilters.forEach((filter: IActiveFilter) => {
+    const valueToLower = filter.value.toLowerCase();
+    filteredAndSortedEditorSteps = filteredAndSortedEditorSteps.filter(
+      (e: Connection) => e.name.toLowerCase().includes(valueToLower)
```
riccardo-forina: Again, we are not dealing with just connections...

riccardo-forina (changes requested):

kahboom:
Okay, I think it's fine now, but anything else, let me know. The specific feedback really helped both in terms of time and getting a better understanding of what I was missing.

Screenshot 2019-06-07 10 35 56

riccardo-forina:
It doesn't build

kahboom:
Sorry I forgot to commit the last change.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2269/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
```diff
@@ -0,0 +1,97 @@
...
+                      <ConnectionSkeleton />
+                    </ConnectionsGridCell>
+                  ))}
+                </>
+              }
+              errorChildren={<ApiError />}
+            >
+              {() => {
+                return this.props.steps.map((s, index) => {
+                  const configurationRequired =
```
gashcrumb: I feel like this could be refactored out into a helper function, probably a function that could take a `Connection | IUIStep`, like:

```
isConfigurationRequired(thingWithBoard: Connection | IUIStep)
```

That way we could have one block of code in the codebase that makes this determination.  I haven't checked but I imagine we could have this pattern in a couple places at this point.

gashcrumb: Other option is to have the function just accept the bulletin board, actually I like that better, i.e.:

```
isConfigurationRequired(board: BulletinBoard)
```

Then you in all cases you could just call `configurationRequired = isConfigurationRequired(s.board);` and be done with it :-)

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
```diff
@@ -0,0 +1,97 @@
...
+                    s.board &&
+                    (s.board!.notices ||
+                      s.board!.warnings ||
+                      s.board!.errors)! > 0;
+
+                  let isTechPreview = false;
+
+                  if (s.connection && s.connection!.connector) {
```
gashcrumb: Similarly with this block.  I'd say it'd be more worthwhile in this case to accept an `IUIStep`, `Connection` and `Connector` object.  Ideally it'd be cool if this could be reduced to:

```
const isTechPreview = figureOutTechPreviewFlag(s);
```

And then of course sort the other spots in the codebase where we're figuring this out.

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorStepsWithToolbar.tsx
```diff
@@ -0,0 +1,117 @@
...
+  extends IEditorStepsProps,
+    Pick<IConnectionsListViewProps, 'createConnectionButtonStyle'> {
+  children?: any;
+}
+
+export class EditorStepsWithToolbar extends React.Component<
+  IEditorStepsWithToolbarProps
+> {
+  public static defaultProps = {
+    includeHidden: false,
```
gashcrumb: I think we want to show the hidden things in this particular case, no?  AFAIK this controls if the API provider connection is displayed and Timer and WebHook.  Actually, since this is now a component dedicated to just this page, I don't think we even need the flag.

gashcrumb: Yeah, the page works as I'd expect, I think this might be a leftover.

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
gashcrumb: BTW I wouldn't apply this refactoring to the integration list if you go this route, that flag shows up under slightly different circumstances.

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorStepsWithToolbar.tsx
kahboom: Yeah this is just left over. Wasn't really looking at it from this perspective tbh, but it makes sense, I'll remove it.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2361/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
kahboom: Just did this but need to go back and refactor the other parts of the app that could use this, I can do it today but I think it should be separate from this PR.

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
kahboom: This would be good, but I think should be addressed in a separate PR.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2366/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
gashcrumb: Why?  You're in the code now, it's probably the best time to tackle this while it's fresh, ya know?  It'd be a matter of writing functions that just return a boolean, then this block here and [this block](https://github.com/syndesisio/syndesis-react/blob/master/app/ui-react/syndesis/src/modules/connections/components/Connections.tsx#L79-L91) will do the same thing then.  Seems like a win to me 'cause you'll be reducing code.

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
gashcrumb: I think it'd be worthwhile to do so now to be honest while you're in there anyways, no?

kahboom:
Changes based on feedback:
- Added `isConfigurationRequired` helper function in `connectionFunctions.ts`. When Riccardo gets [this](https://github.com/syndesisio/syndesis-react/pull/444/files#diff-c8c891bfe81f60cb81839bf4031cb8c7R5) in, we can probably move it there.
- Added `figureOutTechPreviewFlag` helper function in `modules/integrations/editor/utils.ts`. The reason for this, despite that the tech preview flag is used in other places, is that we cannot reference `IUIStep` from the `connectionFunctions.ts` API helpers file, which is defined in the `modules/integrations/editor/interfaces.ts` file.
- Refactored and referenced those functions where possible, which at the moment is limited to `Connections.tsx` and `EditorSteps.tsx`. Didn't do this for `Integrations.tsx` for the reasons mentioned above about the special conditions.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2391/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/integrations/components/editor/utils.ts
```diff
@@ -42,6 +44,44 @@ type StepKindHrefCallback = (
...
+  T extends IUIStep | Connection | Connector
+>(data: T): boolean {
+  if ((data as IUIStep).connection && (data as IUIStep).connection!.connector) {
+    return (
+      getMetadataValue<string>(
+        'tech-preview',
+        (data as IUIStep).connection!.connector!.metadata
+      ) === 'true'
+    );
+  } else if ((data as IUIStep).connector) {
```
gashcrumb: So, you don't probably need this `if` statement, or the next one, at runtime this type info is meaningless 'cause it's gone.

app/ui-react/syndesis/src/modules/integrations/components/editor/utils.ts
```diff
@@ -42,6 +44,44 @@ type StepKindHrefCallback = (
...
 ) => H.LocationDescriptorObject;
 
+/**
+ * Checks whether or not the provided object is a technical preview.
+ * Accepts either an IUIStep, Connection, or a Connector.
+ * Returns a boolean for whether or not the metadata `tech-preview` key
+ * returns a string value of 'true'
+ * @param data
+ */
+export function figureOutTechPreviewFlag<
```
gashcrumb: Probably `isTechPreview` would be better, I know I said `figureOutTechPreviewFlag` but that was being a bit tongue-in-cheek, sorry!.

kahboom: haha no worries, I thought so, but either way we won't be able to use the same declaration so maybe `isTechPreviewCheck` might work better? Or something similar?

app/ui-react/syndesis/src/modules/integrations/components/editor/utils.ts
kahboom: If I do this instead, won't it always return `false` then for whenever the object is a `Connection` or is that a really dumb question? 😖 
```
export function figureOutTechPreviewFlag<
  T extends IUIStep | Connection | Connector
>(data: T): boolean {
  if ((data as IUIStep).connection && (data as IUIStep).connection!.connector) {
    return (
      getMetadataValue<string>(
        'tech-preview',
        (data as IUIStep).connection!.connector!.metadata
      ) === 'true'
    );
  } else if (data as Connector) {
    return getMetadataValue<string>('tech-preview', data.metadata) === 'true';
  } else {
    return false;
  }
}
```

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2395/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/integrations/components/editor/utils.ts
kahboom: I guess my question would be, then why would we want to accept `Connection` as a type? Would it make more sense to only accept `IUIStep` and `Connector`?

kahboom: Hmm I suppose in places that it affects, like the Dashboard Connections list, I can pass a connector instead of connection, but the last question still applies I think. When I remove these `if` statements then the Dashboard Connections no longer displays the flag.

kahboom:
Changes:

- Renamed the tech preview helper function to `isTechPreview`

- Removed the `if` statements as requested, although naturally this means that the Dashboard Connections flag for Tech Preview isn't working as expected (the `if` statements contained the checks that handled these). Please advise.

Screenshot 2019-06-11 09 30 27

_Tech Preview flags no longer displaying for connections on Dashboard_

- I've added a `configRequired` property to `useConnection.tsx`, but until I fully understand what/why I'm doing with respect to the new changes, I don't think it's appropriate to refactor. If this is important then an issue should be raised, as it's not directly related to the bug that this PR is intending to resolve.

kahboom:
In the meantime I've just decided to pass the `connector` from the Connections page (also used for the Dashboard Connections), so that it works again.

Screenshot 2019-06-11 09 47 24

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2411/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
```diff
@@ -78,16 +79,11 @@ export class Connections extends React.Component<IConnectionsProps> {
                               };
 
                               const configurationRequired =
```
riccardo-forina: Because of [this](https://github.com/syndesisio/syndesis-react/pull/409/files#diff-c8c891bfe81f60cb81839bf4031cb8c7R19) you don't need this anymore, `c.configRequired` should hold your value.

app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
```diff
@@ -78,16 +79,11 @@ export class Connections extends React.Component<IConnectionsProps> {
...
-                                  c.board!.warnings ||
-                                  c.board!.errors)! > 0;
+                                c.board && isConfigurationRequired(c.board);
 
-                              const isTechPreview =
-                                c.connector! && c.connector!.metadata!
-                                  ? c.connector!.metadata!['tech-preview'] ===
-                                    'true'
-                                  : false;
+                              const techPreview = c.connector
```
riccardo-forina: Can't this be treated like [this](https://github.com/syndesisio/syndesis-react/pull/409/files#diff-c8c891bfe81f60cb81839bf4031cb8c7R19)?

app/ui-react/packages/api/src/helpers/connectionFunctions.ts
```diff
@@ -82,6 +83,16 @@ export function getActionStepDefinition(
...
 }
 
+/**
+ * Checks whether the ConnectionBulletinBoard provided
+ * @param board
+ */
+export function isConfigurationRequired(
+  board: ConnectionBulletinBoard
+): boolean {
+  return (board!.notices || board!.warnings || board!.errors)! > 0;
```
riccardo-forina: `!`s are useless here, you defined board as a mandatory parameter.

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSteps.tsx
```diff
@@ -0,0 +1,81 @@
...
+                    </ConnectionsGridCell>
+                  ))}
+                </>
+              }
+              errorChildren={<ApiError />}
+            >
+              {() => {
+                return this.props.steps.map((s, index) => {
+                  const configurationRequired =
+                    s.board && isConfigurationRequired(s.board);
```
riccardo-forina: I think this is going to throw for non-connections, since `board` will be `undefined` and `isConfigurationRequired` is assuming board _will_ be defined.

app/ui-react/syndesis/src/modules/integrations/components/editor/utils.ts
```diff
@@ -43,6 +45,30 @@ type StepKindHrefCallback = (
...
 ) => H.LocationDescriptorObject;
 
+/**
+ * Checks whether or not the provided object is a technical preview.
+ * Accepts either an IUIStep, Connection, or a Connector.
+ * Returns a boolean for whether or not the metadata `tech-preview` key
+ * returns a string value of 'true'
+ * @param data
+ */
+export function isTechPreview<T extends IUIStep | Connection | Connector>(
```
riccardo-forina: Do we really need to accept all these types? `Step`s don't have a `connection`, and neither `Extension`s do.

We really just want to accept `Connector` here, since the tech preview metadata is stored there.

Then [this](https://github.com/syndesisio/syndesis-react/pull/409/files#diff-c8c891bfe81f60cb81839bf4031cb8c7R10) can consume this function and set the `isTechPreview` flag.

Bonus points for:
* enhancing the `useConnector` to provide a transformation function (like the `useConnection` does)
* change the `WithConnector` and `WithConnectors` component to use the connector transformation function
* enhance the `transformConnectionResponse` to run the new transformation function on the connection's connector, and retrieve the `isTechPreview` flag from the transformed connector object.

riccardo-forina (changes requested):
I agree with Stan, we either do a full refactoring to smooth all the raw edges, or not.
I'm ok with continuing working on this 'til everything is DRYed up and makes sense.

app/ui-react/syndesis/src/modules/integrations/components/editor/utils.ts
kahboom: >Do we really need to accept all these types? Steps don't have a connection, and neither Extensions do.

That was in reference to this feedback: https://github.com/syndesisio/syndesis-react/pull/409#discussion_r291590677
I'll change it to `Connector` 🤷‍♀

riccardo-forina: Yes, that was a step toward a better refactoring. My suggestion aims to not having to rely on those helpers across the app and enhance the model so that the information is already available for the consumer at all times. If you do what I suggested you can enhance the `IUIStep` interface to also have `isTechPreview` and `isConfigRequired` and you can set those properties to false for steps and extensions [here](https://github.com/syndesisio/syndesis-react/blob/02625588f49dd054095fa8a19d809c099e1aa71f/app/ui-react/syndesis/src/modules/integrations/components/editor/utils.ts#L57) where the `IUIStep` object is generated.

app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
kahboom: >I agree with Stan, we either do a full refactoring to smooth all the raw edges, or not.
>I'm ok with continuing working on this 'til everything is DRYed up and makes sense.

As long as you're ok with questions and helping to explain, I'm always happy to refactor.

`configRequired` is not a property of `ConnectionOverview`, will run into this for `isTechPreview` as well, I'm guessing--what is the best approach here? Should I make an interface that extends `ConnectionOverview`?

app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
kahboom: `techPreview` relied on `IUIStep` based on the feedback from before, which couldn't be access from those API helper files since it was within `modules/integrations/components/editor/interfaces.ts`. If we only accept `Connector` then it should be fine and I can get rid of the generics stuff as well.

app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
riccardo-forina: I think we should be fine without a new interface. When you get connections out from `useConnections` (or `WithConnections`) the object that's returned is an object that does respect the `Connection` interface, but it's also extended with a number of properties that the app needs.

These connection objects are then eventually passed to the `toUIStep` function to convert them, and all their properties are spread in the new `IUIStep` one.
So if we simply extend the `IUIStep` interface to have `configRequired` and `isTechPreview` defined, we should be good.

kahboom: I'm able to do this for `EditorSteps.tsx`, for instance, but not for `Connections.tsx` since it seems to never use `IUIStep`, where in `IConnectionsProps`: `connections: ConnectionOverview[];`. Should that instead be `IUIStep` or have I missed something?

riccardo-forina: `Connetions.tsx` will use connections coming directly from the API, so already transformed. `c.configRequired` will be available then.

riccardo-forina: But I see now that we might need to create an extended `IConnection` interface, since `Connections.tsx`. Boring, but yeah, needed.

app/ui-react/syndesis/src/modules/integrations/components/editor/utils.ts
kahboom: I'm following but am not sure if I'm doing this properly, please see my changes. Also, was easier to start with `WithConnectors.tsx` first.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2439/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
```diff
@@ -13,11 +13,17 @@ import { Translation } from 'react-i18next';
 import { UIContext } from '../../../app';
 import { ApiError, EntityIcon } from '../../../shared';
 
+export interface IConnectionOverview extends ConnectionOverview {
```
riccardo-forina: Yes something like this, only this should be in the `extra.d.ts` and should be referenced by the `useConnection`, etc. components/hooks to explain their return type.

Oh and no need to specify `board`, it's already part of the `ConnectionOverview` interface.

riccardo-forina (commented):
Check the comment up here

app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
kahboom: I pushed up some changes, it's a bit difficult to test with staging being down. I moved the interface and the `board` prop, updated the reference in `useConnection` and `Connections.tsx`, but if I missed anything let me know.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2453/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2482/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2565/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/packages/api/src/useConnection.tsx
```diff
@@ -1,8 +1,16 @@
...
+  isDerived,
+  isTechPreview,
+} from './helpers';
 import { useApiResource } from './useApiResource';
 
-export const transformConnectionResponse = (connection: ConnectionOverview) => {
+export const transformConnectionResponse = (
+  connection: IConnectionOverview
+) => {
   return {
```
riccardo-forina, line 17: I think here we should get the updated connector like `const connector = transformConnectorResponse(connection.connector!);` and use it after.
`isTechPreview` would become `isTechPreview: connector.isTechPreview`

app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
```diff
@@ -17,7 +17,7 @@ export interface IConnectionsProps {
   error: boolean;
   includeConnectionMenu: boolean;
   loading: boolean;
-  connections: ConnectionOverview[];
+  connections: IConnectionOverview[];
 
   getConnectionHref(connection: ConnectionOverview): H.LocationDescriptor;
```
riccardo-forina: can we upgrade this to `IConnectionOverview`?

app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
```diff
@@ -17,7 +17,7 @@ export interface IConnectionsProps {
...
   getConnectionEditHref?(connection: ConnectionOverview): H.LocationDescriptor;
```
riccardo-forina: can we upgrade this to `IConnectionOverview`?

app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
```diff
@@ -77,25 +77,11 @@ export class Connections extends React.Component<IConnectionsProps> {
...
-                                  : false;
-
                               return (
                                 <ConnectionsGridCell key={index}>
                                   <ConnectionCard
                                     name={c.name}
-                                    configurationRequired={
-                                      configurationRequired
-                                    }
+                                    configurationRequired={c.configRequired}
```
riccardo-forina: 🎉

app/ui-react/syndesis/src/modules/connections/components/Connections.tsx
```diff
@@ -135,7 +121,7 @@ export class Connections extends React.Component<IConnectionsProps> {
                                           }
                                         : undefined
                                     }
-                                    techPreview={isTechPreview}
+                                    techPreview={c.isTechPreview}
```
riccardo-forina: 🥳

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorStepsWithToolbar.tsx
```diff
@@ -0,0 +1,113 @@
...
+  id: 'name',
+  isNumeric: false,
+  title: i18n.t('shared:Name'),
+} as ISortType;
+
+const sortTypes: ISortType[] = [sortByName];
+
+export interface IEditorStepsWithToolbarProps
+  extends IEditorStepsProps,
+    Pick<IConnectionsListViewProps, 'createConnectionButtonStyle'> {
```
riccardo-forina, line 60: I don't think this `Pick` is needed at all

app/ui-react/syndesis/src/modules/integrations/components/editor/interfaces.ts
```diff
@@ -29,7 +29,9 @@ export enum DataShapeDirection {
 
 export interface IUIStep extends StepKind {
   board?: ConnectionBulletinBoard;
+  configRequired?: boolean;
```
riccardo-forina: I'd make this required, just to stay on the safe side and highlight any point where we missed passing the updated structure.

app/ui-react/syndesis/src/modules/integrations/components/editor/interfaces.ts
```diff
@@ -29,7 +29,9 @@ export enum DataShapeDirection {
...
   connector?: Connector;
+  isTechPreview?: boolean;
```
riccardo-forina: I'd make this required, just to stay on the safe side and highlight any point where we missed passing the updated structure.

riccardo-forina (changes requested):
Great progress, this is almost done! I left a few comments :P

PR #418, opened 2019-06-05
Fix for Integration Editor Action ListItem row

### LINKED ISSUE #417: 🛠 ListViews issues (actions)
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description

The layout of the actions choice for a connector is quite different. It's related to issues with the list view, so it might actually appear in multiple places

React

Screenshot from 2019-06-05 13-48-40

Angular

Screenshot from 2019-06-05 13-55-44

kahboom:
Screenshot 2019-06-05 13 17 33

fixes #417

### DIFF
--- a/app/ui-react/packages/ui/src/Integration/Editor/IntegrationEditorActionsListItem.tsx
+++ b/app/ui-react/packages/ui/src/Integration/Editor/IntegrationEditorActionsListItem.tsx
@@ -19,8 +19,8 @@ export class IntegrationEditorActionsListItem extends React.Component<
         )}-list-item`}
         actions={this.props.actions}
         heading={this.props.integrationName}
-        description={this.props.integrationDescription}
-        stacked={true}
+        description={<b>{this.props.integrationDescription}</b>}
+        stacked={false}
       />
     );
   }
--- a/app/ui-react/packages/ui/src/Integration/Editor/IntegrationEditorActionsListItem.tsx
+++ b/app/ui-react/packages/ui/src/Integration/Editor/IntegrationEditorActionsListItem.tsx
@@ -18,8 +18,12 @@ export class IntegrationEditorActionsListItem extends React.Component<
           this.props.integrationName
         )}-list-item`}
         actions={this.props.actions}
-        heading={this.props.integrationName}
-        description={<b>{this.props.integrationDescription}</b>}
+        heading={<b>{this.props.integrationName}</b>}
+        description={
+          this.props.integrationName === this.props.integrationDescription
+            ? undefined
+            : this.props.integrationDescription
+        }
         stacked={false}
       />
     );


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2190/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

paoloantinori:
> The fact that there's no real description for these items and they only repeat the title makes it a little more of an eye sore

You are generalizing over a specific example.

gaughan:
I think the description item has come up a few times. It would be great to have the expanded menu offer a more rich feel. Maybe we can look into this when we do a 7.4 review

dongniwang:
I think a typical list row contains list item name, list item description, and we might bold the list item name in some cases, but never really bolding the description. The order we have in the screenshot gives me an odd feeling. 

I'm not sure if right now we want to align all the bold styling across all lists in the app...and I do have an issue open https://github.com/syndesisio/syndesis/issues/5525 to investigate list behavior. 

@gaughan - can you share more as to what do you mean by expanded menu and rich feel? 
> It would be great to have the expanded menu offer a more rich feel.

gaughan:
Apologies, disregard, I was cross-posting. 
I do agree, the duplication of text "cron timer" cron timer" should be avoided tho.
Also, can we just call it timer?

seanforyou23:
Sorry wasn't trying to overgeneralize the problem @paoloantinori - just didn't want to push back on the bolding without offering some alternate paths forward and I saw the stub description text as further complicating the decision. If the description is the same as the title text and we don't bold one of them then it looks like a bug again, even with them not being stacked. It's all a consistency game from my point of view and as @dongniwang mentioned, if we were to bold anything, the title would make most sense and not the description.

I dunno about using only "Timer" as a title, as there a multiple types of timers. I'd think to keep the titles as is and just change the descriptions to something that points out that these actions are _types_ of timers. Maybe time based vs interval based would be a good thing to point out in the descriptions?

seanforyou23:
But that's ok I would think because this is what the description is here to help with - and do we expect users to know what every action is just by the name? A user may not know what cron is, but if we describe it as a "type of timer that schedules jobs to run periodically at fixed times", then they may understand? I just don't want to undercut the importance of using and relying on sound descriptions.

gaughan:
I agree on solid descriptions. But it's a backup, user don't read it until something goes wrong.

dongniwang:
Here's what I think we can do right now:
1. Keep the current layout. I'm fine with the stacked if that introduces less hassle. If we want to change it back to the horizontal view, that'd be great as well. 
2. If we go with the horizontal view, let's bold the name and unbold the description. That would help differentiate these two kinds of content.  
3. I don't think introducing name changes right now is a good idea. And it deserves a separate discussion. If we want to change anything or to avoid getting into a situation where we have duplicating text, we should be focusing on having a more descriptive description. Yes, users may not read until something goes wrong, but this should not be the reason for not provide them with sufficient information IMO.

kahboom:
For comparison sake, here is what it looks like when not bolded. Less of an eye sore, but slightly more confusing:

Screenshot 2019-06-05 13 18 00

We probably just need to get the descriptions changed, as you guys suggested. For the purposes of this PR, however, should we bold or no bold?

gashcrumb:
Yeah, so I'd get rid of the bold and also to prevent the duplicates I'd definitely compare the name/description and just show the name and hide the description (replace it with an `nbsp;`) if it's exactly the same as the description.  At least we'll kinda look like we know what we're doing.

riccardo-forina:
Better pass undefined to the description and set stacked to false to ensure vertical alignment of the name when it’s alone in the row

kahboom:
Based on the feedback:
- Per @dongniwang , no name changes, only bold the name and not the description.
- Per @gashcrumb , compare name to description to determine whether or not to show the description.
- Per @riccardo-forina , returned `undefined` for description if the name and description match.

See screenshot below:

Screenshot 2019-06-06 12 23 32

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2236/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

dongniwang:
Thanks @kahboom for making the update! Looks good to me. 

I do want to reiterate that I think having a more descriptive description is the way to go.

gashcrumb:
> I do want to reiterate that I think having a more descriptive description is the way to go

@dongniwang can you log a bug for that in the main syndesis repo?  That's a matter of updating the connector definition, those names/descriptions don't come from the UI at all.

gashcrumb:
Make it specifically for the timer connection BTW.

dongniwang:
https://github.com/syndesisio/syndesis/issues/5625 here we go!

riccardo-forina:
Although if we really want to be thorough, all the text coming from the BE should be matched against some i18n dictionary so to translate also those bits.

Not that we have any real support for anything but EN, but it would allow us to make UX improvements without touching the BE at all...

gashcrumb:
> Although if we really want to be thorough, all the text coming from the BE should be matched against some i18n dictionary so to translate also those bits.

I absolutely agree but then we've just increased the scope of solving the problem.  For 7.4 let's keep it to updating the connector definition files, and absolutely we should plan on generally avoiding using text from the BE for anything and instead have it use i18n keys.

riccardo-forina:
Of course! That was definitely to be filed as an improvement thing :D

PR #424, opened 2019-06-05
Fixes an issue when deleting an integration from it's detail page

elvisisking:
- see issue syndesisio/syndesis#5578
- added WithRouter to WithIntegrationActions
- after a successful delete, the integrations list page is displayed

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/components/WithIntegrationActions.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/WithIntegrationActions.tsx
@@ -4,6 +4,7 @@ import {
   canEdit,
   WithIntegrationHelpers,
 } from '@syndesis/api';
+import * as H from '@syndesis/history';
 import { IntegrationOverview } from '@syndesis/models';
 import {
   ConfirmationButtonStyle,
@@ -31,6 +32,7 @@ export interface IWithIntegrationActionsChildrenProps {
 
 export interface IWithIntegrationActionsProps {
   integration: IntegrationOverview;
+  postDeleteHref?: H.LocationDescriptorObject;
   children: (props: IWithIntegrationActionsChildrenProps) => any;
 }
 
@@ -234,7 +236,11 @@ export class WithIntegrationActions extends React.Component<
                                   await deleteIntegration(
                                     this.props.integration.id!
                                   );
-                                  history.push(resolvers.list());
+
+                                  // redirect if requested
+                                  if (this.props.postDeleteHref) {
+                                    history.push(this.props.postDeleteHref);
+                                  }
                                 } catch (err) {
                                   pushNotification(
                                     i18n.t(
--- a/app/ui-react/syndesis/src/modules/integrations/components/WithIntegrationActions.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/WithIntegrationActions.tsx
@@ -11,6 +11,7 @@ import {
   ConfirmationIconType,
   IIntegrationAction,
 } from '@syndesis/ui';
+import { WithRouter } from '@syndesis/utils';
 import * as React from 'react';
 import { Translation } from 'react-i18next';
 import { UIContext } from '../../../app';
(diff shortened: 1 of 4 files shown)


riccardo-forina:
What if I delete the integration from the dashboard? Will it go to the integrations list?

elvisisking:
It does now go to the list page @riccardo-forina:( I need to fix that.

riccardo-forina:
I dunno it’s not that big of a deal I think. @dongniwang what do you think? Is it a problem if users are redirected to the integrations list page after deleting an integration, regardless of where they were? Eg. deleting from the dashboard will take you to the integrations list page.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2215/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

gashcrumb:
I think the redirect should only occur if you're on the detail page.  It'd definitely be weird to delete an integration on the dashboard there and then wind up on the integration list page.

dongniwang:
Yeah, I think it makes sense to delete and return to the integration list when you're on the details page. 

But if delete happens from the dashboard, I would expect to see a confirmation dialog then a toast notification saying that the delete was successful. 

We (meaning UX) should be better in documenting these interactions. Thanks for fixing the behavior! And I think it's not a big deal that it takes users back to the list page.

elvisisking:
@riccardo-forina @gashcrumb wdyt of this code that conditionally redirects to the integrations list page only when on the integration details page:
```
if (history.location.pathname.endsWith("details")) {
  history.push(resolvers.list());
}
``` 
And might there be a better way to do this?

riccardo-forina:
Yeah, please no :D

What about adding a `postDeleteHref: H.LocationDescriptionObject` property to the `WithIntegrationActions` component, and let the consumers of that component specify the right URL to use for the redirect?

elvisisking:
Yeah that is much nicer @riccardo-forina . Will do.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2322/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

elvisisking:
While things are working as expected when an integration is deleted (redirect to list when on details page, otherwise stay on same page), I'm getting the following error in the console when the delete action from the kebab menu on the integration `DetailsPage` and `MetricsPage` is used:
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in Fetch (created by Context.Consumer)
    in SyndesisFetch (created by WithChangeListener)
    in WithChangeListener (created by Context.Consumer)
    in Fetch (created by Context.Consumer)
    in SyndesisFetch (created by WithIntegration)
    in WithIntegration (created by WithMonitoredIntegration)
    in WithMonitoredIntegration (at DetailsPage.tsx:46)
    in Route (created by WithRouteData)
    in WithRouteData (at DetailsPage.tsx:44)
    in Translation (at DetailsPage.tsx:40)
    in DetailsPage (created by Context.Consumer)
```
It seems like the code in [`ChangeListener.componentWillUnmount()`](https://github.com/syndesisio/syndesis/blob/master/app/ui-react/packages/api/src/WithChangeListener.tsx#L48) should not let this happen but not sure. I've tried a few things in the details page to fix this but seems like this could be a bigger issue especially when deleting objects from detail pages.

gashcrumb:
Raise an issue for that and I'll investigate next week, shouldn't be a
problem for getting this pr merged I think.  Unless you want to look into
it further that is.

riccardo-forina:
Ooooh it was the Fetch raising the warning, you didn’t t mention that! That’s the `fetch` promise that will still trigger even if the `Fetch` component has been unmounted. The problem is that no browser implements abortable fetches, so there is no way to work around this without replacing the browser’s own `fetch` with some polyfilled version that’s abortable, like `whatwg-fetch` (which is basically an `xhr` request wrapped in a `fetch` compatible interface). Or drop it in favor of something like `axios`.

Anyway, this isn’t something urgent, the warning is safe to ignore in this case (but should be ultimately addressed)

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2370/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2380/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

paoloantinori:
@riccardo-forina with your last comment, are you saying that we can merge this?

PR #426, opened 2019-06-06
feat: oauth flow for connections

riccardo-forina:
This is implemented up to the redirect back from the server.

## Changes
oauth

## Todo

* [x] proper layout for the "Connect" page
* [x] pass the state to the review page (probably as configurationProperties, still have to check this)
* [ ] the connection details page

## Issues
* [x] the connect thing is not working for me at the moment, I'm getting a "timeout" error of some sort from the BE (minishift)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2238/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

zregvart:
This changes the OAuth login to open in a pop-up, should we worry about pop-up blockers?

riccardo-forina:
The eventuality is trapped and a relative toast is displayed to the user explaining what’s happening. That said, we are opening a pop up as a direct and sole consequence of an user action, pointing to well known domains. I think the risk is quite remote.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2273/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2292/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

riccardo-forina:
@dongniwang the OAuth flow is slightly different than before, instead of redirecting the user to the 3rd party to eventually get back to Syndesis, the 3rd party authorization will happen in a popup. Advantages:
* if the connector is not configured correctly, the user will remain in Syndesis and only the popup will show the misconfiguration error. Before you would have seen the 3rd party error page - which is often pretty bad - without any way to get back to Syndesis but the back button.
* this makes the process a bit quicker because we don't have to reload the whole app.
* technically, it's easier and cleaner

riccardo-forina:
This is the error condition I'm talking about.

oauth-misconfigured-angular

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2297/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

riccardo-forina:
I think this can be reviewed. Better handle the details page in another PR so to enable QA to do their thing.

dongniwang:
Thanks for the update @riccardo-forina !

- I agreed that having an error on a 3rd party page and no way to return back to Syndesis is bad...
- Good to know we have the ability to intercept and display a toast notification. 
- What would happen when OAuth is successfully done, would the pop-up remain opened? What would be shown in the pop-up at that point?
- Does the pop-up apply to all the connectors we have for the create connection flow? Or is it only for connectors like Salesforce?

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2300/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

riccardo-forina:
Check the gif in the description, you can see the pop up there. Anyway, it will show a link asking to close the window, and that it will close automatically after 5s. And no, popup and messages are all the same for all the oauth connectors

dongniwang:
What would happen if the OAuth didn't go as planned? Do we have a toast notification or inline notification - assuming users would stay on step 2 - for that?

riccardo-forina:
It depends on when the flow broke, if it was on the 3rd party page we have no control over that, it it was on the end of the flow the popup will close and a toast error notification with the error message will be displayed

zregvart:
Epic work, thank you @riccardo-forina :+1:

app/ui-react/syndesis/public/oauth-redirect.html
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
+<script>
+  let autocloseInterval;
+
+  function authCompleted() {
+    clearInterval(autocloseInterval);
+    const cred = document.cookie.split(';').filter(c => c.indexOf('cred-o2') === 0)[0];
```
zregvart: This might fail for OAuth1 (Twitter) the cookie is named `cred-o1` there. Also there could be multiple `cred-*` cookies if the user is trying multiple times to connect, picking the first one might prove to be a bit flaky.

I still can't wrap my head around the need for this, even seeing it with my own eyes that the updated cookie sent from `.../callback` is reset to the old value as soon as the popup is closed.

riccardo-forina: Thanks for the o1/o2 thing, I wasn’t aware. There is no risk for multiple cookies because we clean ‘em all when you enter the configuration page

riccardo-forina: And yeah, cookies and SPAs and html5 history is a dreadful combo

Issue #425, opened 2019-06-05
Clicking on mobile wizard toggle icon doesn't work


mcoker:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
If you click on the down arrow icon, the wizard menu doesn't toggle on mobile. It toggles if you click anywhere else in that area though.

## Expected behavior
Clicking anywhere on the toggle should toggle the wizard step menu.

## Screenshot
<img width="602" alt="Screen Shot 2019-06-05 at 4 41 48 PM" src="https://user-images.githubusercontent.com/35148959/58992667-12730600-87b1-11e9-87b7-ca20dd474b79.png">

## Tasks involved / Steps to Reproduce

1. Create connection or api client connector
2. shrink viewport
3. click on the down arrow in the wizard steps menu
4. 😢

PR #431, opened 2019-06-06
fix(wiz-toggle): fix wizard toggle icon click-ability

seanforyou23:
This PR updates the header to more seamlessly collapse/consolidate the user menu at smaller viewports. Previously we would collapse logout into the help menu (keeping the help icon) and later convert that icon to kebab. After further reviewing this interaction, we realized it seems strange and a better experience is to convert the icon to a kebab at the same time we move/consolidate the two menus.

Also fixes a bug where selecting the collapse "toggle" icon for wizard  wasn't actually collapsing the menu. Now it does.

One question that came up along the way was "how are we achieving this collapse functionality for the PF3 Wiz steps?". I don't see this capability in the PF3 react implementation;

http://patternfly-react.surge.sh/patternfly-3/index.html?selectedKind=patternfly-react%2FCommunication%2FWizard%2FComponents&selectedStory=Wizard&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

But it does seem to be a thing in core;

https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/wizard.html

Should help close https://github.com/syndesisio/syndesis-react/issues/425

allow pointer events to pass through for wiz toggle icon
introduce tabletView mode where we can update aspects of the UI based on viewport size
update header icons depending on viewport size

### DIFF
--- a/app/ui-react/packages/ui/src/Layout/AppLayout.tsx
+++ b/app/ui-react/packages/ui/src/Layout/AppLayout.tsx
@@ -14,6 +14,7 @@ import {
   ToolbarItem,
 } from '@patternfly/react-core';
 import { css } from '@patternfly/react-styles';
+import { global_breakpoint_lg } from '@patternfly/react-tokens';
 import * as React from 'react';
 import { HelpDropdown } from '../Shared/HelpDropdown';
 import { AppLayoutContext } from './AppLayoutContext';
@@ -83,8 +84,15 @@ export const AppLayout: React.FunctionComponent<ILayoutBase> = ({
     setCurViewportWidth(props.windowSize);
   };
   const [isTabletView, setIsTabletView] = React.useState(false);
+  const LARGE_VIEWPORT_BREAKPOINT = parseInt(
+    global_breakpoint_lg.value.substring(
+      0,
+      global_breakpoint_lg.value.indexOf('px')
+    ),
+    10
+  );
   React.useEffect(() => {
-    setIsTabletView(curViewportWidth <= 992);
+    setIsTabletView(curViewportWidth <= LARGE_VIEWPORT_BREAKPOINT);
   }, [curViewportWidth]);
   return (
     <AppLayoutContext.Provider
--- a/app/ui-react/packages/ui/src/Layout/AppLayout.tsx
+++ b/app/ui-react/packages/ui/src/Layout/AppLayout.tsx
@@ -77,9 +77,15 @@ export const AppLayout: React.FunctionComponent<ILayoutBase> = ({
   const onNavToggleMobile = () => {
     setIsNavOpenMobile(!isNavOpenMobile);
   };
+  const [curViewportWidth, setCurViewportWidth] = React.useState(1024);
   const onPageResize = (props: { mobileView: boolean; windowSize: number }) => {
     setIsMobileView(props.mobileView);
+    setCurViewportWidth(props.windowSize);
   };
+  const [isTabletView, setIsTabletView] = React.useState(false);
+  React.useEffect(() => {
+    setIsTabletView(curViewportWidth <= 992);
+  }, [curViewportWidth]);
   return (
     <AppLayoutContext.Provider
       value={{
@@ -103,7 +109,7 @@ export const AppLayout: React.FunctionComponent<ILayoutBase> = ({
                 >
                   <ToolbarItem>
                     <HelpDropdown
-                      isMobileView={isMobileView}
+                      isTabletView={isTabletView}
                       className="syn-help-dropdown"
                       isOpen={false}
                       launchSupportPage={onSelectSupport}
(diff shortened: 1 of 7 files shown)


mcoker:
looks like you have a build error

```
ERR! => Failed to build the preview
ERR! [at-loader] ./stories/Shared/HelpDropdown.stories.tsx:13:10 
ERR!     TS2322: Type '{ dropdownDirection: any; dropdownPosition: any; isMobileView: any; launchConnectorsGuide: any; launchContactUs: any; launchSampleIntegrationTutorials: any; launchSupportPage: any; launchUserGuide: any; launchAboutModal: any; isOpen: true; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<HelpDropdown> & Readonly<IHelpDropdownProps> & Readonly<{ children?: ReactNode; }>'.
ERR!   Property 'isMobileView' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<HelpDropdown> & Readonly<IHelpDropdownProps> & Readonly<{ children?: ReactNode; }>'.
(node:2332) UnhandledPromiseRejectionWarning: [object Object]
(node:2332) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:2332) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
error Command failed with exit code 1.
```

app/ui-react/packages/ui/src/Layout/AppLayout.tsx
```diff
@@ -77,9 +77,15 @@ export const AppLayout: React.FunctionComponent<ILayoutBase> = ({
...
     setIsNavOpenMobile(!isNavOpenMobile);
   };
+  const [curViewportWidth, setCurViewportWidth] = React.useState(1024);
   const onPageResize = (props: { mobileView: boolean; windowSize: number }) => {
     setIsMobileView(props.mobileView);
+    setCurViewportWidth(props.windowSize);
   };
+  const [isTabletView, setIsTabletView] = React.useState(false);
+  React.useEffect(() => {
+    setIsTabletView(curViewportWidth <= 992);
```
gashcrumb: What's this number?  Seems kinda arbitrary.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2284/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/packages/ui/src/Layout/AppLayout.tsx
seanforyou23: Ah yea so it's actually a common breakpoint, we use it in core for denoting a "large" viewport - https://github.com/patternfly/patternfly-next/blob/master/src/patternfly/sass-utilities/scss-variables.scss#L154

I'm looking now to see if we can pull this value dynamically from one of the styling packages.

gashcrumb: TBH even just naming it with a const would be good like `const SUPER_IMPORTANT_BREAKPOINT=992` and then use that, then it's kinda self documenting :-)

mcoker:
LGTM!

app/ui-react/packages/ui/src/Layout/AppLayout.tsx
seanforyou23: Can do!

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2296/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/packages/ui/src/Layout/AppLayout.tsx
```diff
@@ -77,9 +78,22 @@ export const AppLayout: React.FunctionComponent<ILayoutBase> = ({
...
+  const LARGE_VIEWPORT_BREAKPOINT = parseInt(
+    global_breakpoint_lg.value.substring(
+      0,
+      global_breakpoint_lg.value.indexOf('px')
+    ),
+    10
+  );
+  React.useEffect(() => {
+    setIsTabletView(curViewportWidth <= LARGE_VIEWPORT_BREAKPOINT);
+  }, [curViewportWidth]);
```
riccardo-forina: You're missing the `setIsTabletView` function in the list of dependencies

seanforyou23: Not sure I understand, what do you mean missing? It's a local setter that doesn't need to be driven based on anything other than the viewport size.. as far as I can tell anyway. Is there a bug I'm not seeing?

riccardo-forina: Your effect depends on `setIsTabletView`, so it should be declared in the list of dependencies together with `curViewportWidth`.

It's a _potential source_ of bug. In this case it wouldn't matter, but better start making an eye for these things. We don't have [this](https://github.com/Gelio/tslint-react-hooks) for the packages yet, only for the main app; that would have warned you about this.

PR #434, opened 2019-06-06
Fixes for query client issues

mdrillin:
Fixes for ENTESB-10792
- add title and row count above results table - SqlClientContext
- WithVirtualizationSqlClientForm - fixes row and column handling when there are no rows returned.
- VirtualizationUtils now escapes preview query names in the event that the connection has a special character
- fix issue from previous commit - additionalInfo in VirtualiationListItem was not specified correctly

### DIFF
--- a/app/ui-react/syndesis/src/modules/data/shared/VirtualizationUtils.ts
+++ b/app/ui-react/syndesis/src/modules/data/shared/VirtualizationUtils.ts
@@ -417,10 +417,13 @@ export function getPreviewSql(viewDefinition: ViewDefinition): string {
 function getPreviewTableName(sourcePath: string): string {
   // Assemble the name, utilizing the schema model suffix
   return (
+    '"' +
     getConnectionName(sourcePath).toLowerCase() +
     SCHEMA_MODEL_SUFFIX +
-    '.' +
-    getNodeName(sourcePath)
+    '".' +
+    '"' +
+    getNodeName(sourcePath) +
+    '"'
   );
 }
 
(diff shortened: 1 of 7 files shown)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2264/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/data/shared/VirtualizationUtils.ts
```diff
@@ -417,10 +417,13 @@ export function getPreviewSql(viewDefinition: ViewDefinition): string {
...
   return (
+    '"' +
     getConnectionName(sourcePath).toLowerCase() +
     SCHEMA_MODEL_SUFFIX +
-    '.' +
-    getNodeName(sourcePath)
+    '".' +
+    '"' +
+    getNodeName(sourcePath) +
+    '"'
```
gashcrumb: You could do:

```
`"${getConnectionNAme(sourcePath).toLowerCase()}${SCHEMA_MODEL_SUFFIX}.${getNodeName(sourcePath)}"."${getNodeName(sourcePath)"`
```
Using a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) if you'd rather avoid all the `+` etc.  Depends on what you think is easier to read, I've been finding though that the template literals make me more comfortable to use than building complex strings like this.

mdrillin: Yeah, agree it would be easier to read.  I can look at changing in next commit. Thx!

PR #436, opened 2019-06-07
fix: Add syntax examples in condition help block

christophd:
Fixes #5485

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/locales/integrations-translations.en.json
+++ b/app/ui-react/syndesis/src/modules/integrations/locales/integrations-translations.en.json
@@ -81,9 +81,9 @@
       "saveAndPublish": "Save and publish"
     },
     "choiceForm": {
-      "conditionDescription": "Provide a condition that you want to evaluate.",
+      "conditionDescription": "Provide a condition that you want to evaluate (e.g. ${in.header.type} == 'note' or ${in.body.title} contains 'Important').",
       "conditionName": "Condition",
-      "conditionPlaceholder": "Condition",
+      "conditionPlaceholder": "Simple language expression",
       "addCondition": "+ Add another condition",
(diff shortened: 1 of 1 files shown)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2282/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

gashcrumb:
Thanks for the PR!  I wonder though, the text looks a bit repetitive in practice:

image

Could add a `<p>` [here](https://github.com/syndesisio/syndesis-react/blob/master/app/ui-react/syndesis/src/modules/integrations/components/editor/choice/ChoiceStepPage.tsx#L209) with the syntax examples.  I wonder then if we'd even need the field level help block.

gashcrumb:
@dongniwang thoughts on that actually?  Or would we rather have the field-level help like that?

christophd:
@gashcrumb I tried this with a wider screen so that the help blocks were not using a line breaks. In your screenshot the repeating help blocks do look a bit overloaded, agreed

gashcrumb:
Yeah, you're right that does look better.  Lately I've been thinking I might need a monitor upgrade :-)

dongniwang:
Awesome! Thanks for the PR! It's a great step forward now that we provide more help to our users. 

In the screenshot - What are the texts inside the input fields? Are they prefill by default? Feel like the screen is a little busy with all the texts and users might lose focus on what's the primary task here on the screen.

dongniwang:
Ran into this strange behavior when looking at the PR. 

Seems like we replace the helper text below the input field with validation text in a pretty subtle way. And I can't seem to be able to get the helper text to show up again. See screenshot below. 

cc: @seanforyou23 here for his input as well. 

conditionalFlow-helptext

PR #439, opened 2019-06-07
chore: Data Mapper should skip maven classpath resolution in Syndesis…

igarashitm:
… (syndesisio/syndesis#5622)

A bit better error message for syndesisio/syndesis#5622, show a Step ID for which Document fails to load. atlasmap/atlasmap#971 will add human readable Document name in addition.

### DIFF
--- a/app/ui-react/packages/atlasmap-assembly/src/environments/environment.ts
+++ b/app/ui-react/packages/atlasmap-assembly/src/environments/environment.ts
@@ -3,6 +3,8 @@
 // The list of file replacements can be found in `angular.json`.
 
 export const environment = {
+  // AtlasMap skips Maven classpath resolution if (classpath)
+  classpath: ' ',
   production: false,
   xsrf: {
     headerName: 'SYNDESIS-XSRF-TOKEN',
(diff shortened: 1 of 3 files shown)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2313/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/packages/atlasmap-assembly/src/environments/environment.ts
```diff
@@ -3,6 +3,7 @@
 // The list of file replacements can be found in `angular.json`.
 
 export const environment = {
+  classpath: ' ',
```
gashcrumb: Is the space significant at all?  If so could you maybe pop some comments in here as to why?

gashcrumb: mostly because as @riccardo-forina has figured out, I'm extremely forgetful 😄

igarashitm: ah good point, I'll add it - this is because AtlasMap triggers if (!classpath), and even empty classpath without space enter into that if clause. Yep, kinda dirty hack. I think Syndesis will never use maven classpath resolution though. It's incredibly slow.

igarashitm: OK done

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2323/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #443, opened 2019-06-10
fix: template step requires an output data shape to be selected

riccardo-forina:
Bug reported [here](https://github.com/syndesisio/syndesis/issues/5622#issuecomment-500343788).

(oh also fixes a bug with the integration list item that was causing a lot of console errors to be logged)

### DIFF
--- a/app/ui-react/packages/api/src/WithSteps.tsx
+++ b/app/ui-react/packages/api/src/WithSteps.tsx
@@ -48,7 +48,7 @@ export const ALL_STEPS: StepKind[] = [
     },
     true
   ),
-  {
+  requiresOutputDataShape({
     id: undefined,
     connection: undefined,
     action: undefined,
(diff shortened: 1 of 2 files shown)


riccardo-forina:
Fixes [https://github.com/syndesisio/syndesis/issues/5652]

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2358/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

gashcrumb:
So, we've actually relaxed some restrictions beyond the angular app now in the react app.  I think we decided that the template step can just be available whenever, right @christophd ?

phantomjinx:
> So, we've actually relaxed some restrictions beyond the angular app now in the react app. I think we decided that the template step can just be available whenever, right @christophd ?

Given the template-step requires an input datashape as the datasource of its symbol substitution, I can't understand how that could be the case, eg.

- Timer -> Template -> Log
- Template's template: The Quick ${brown} fox jumped over the lazy dog
- Where does the data come from that ${brown} gets substituted for?

gashcrumb:
Ah, I see, ok great, thanks!

PR #445, opened 2019-06-10
fix(5273): Mustache errors showing wrong line number

phantomjinx:
* CodeMirror editor starts at line 0 (due to array index) so need to
  increment in order to display correct line number to user.

* This only affects the mustache lint as both velocity and freemarker
  already deal with this internally.


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2372/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #447, opened 2019-06-10
TTOOLS-651 Clear query results on form change

mdrillin:
Added validate method on the SqlClient form.  Its purpose is to reset the query results table upon change of one of the controls.

### DIFF
--- a/app/ui-react/syndesis/src/modules/data/shared/WithVirtualizationSqlClientForm.tsx
+++ b/app/ui-react/syndesis/src/modules/data/shared/WithVirtualizationSqlClientForm.tsx
@@ -1,5 +1,5 @@
 import { WithVirtualizationHelpers } from '@syndesis/api';
-import { AutoForm, IFormDefinition } from '@syndesis/auto-form';
+import { AutoForm, IFormDefinition, IFormValue } from '@syndesis/auto-form';
 import * as H from '@syndesis/history';
 import { QueryResults, ViewDefinition } from '@syndesis/models';
 import { SqlClientContent, SqlClientForm } from '@syndesis/ui';
@@ -194,6 +194,13 @@ export class WithVirtualizationSqlClientForm extends React.Component<
       view: this.getInitialView(),
     };
 
+    // The purpose of this function is to reset the query results
+    //  whenever a form selection is changed
+    const validate = (values: IFormValue) => {
+      this.setQueryResults(WithVirtualizationSqlClientForm.queryResultsEmpty);
+      return {};
+    };
+
     return (
       <Translation ns={['data', 'shared']}>
         {t => (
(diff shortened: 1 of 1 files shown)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2394/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/syndesis/src/modules/data/shared/WithVirtualizationSqlClientForm.tsx
```diff
@@ -194,6 +194,11 @@ export class WithVirtualizationSqlClientForm extends React.Component<
       view: this.getInitialView(),
     };
 
+    const validate = (values: IFormValue) => {
+      this.setQueryResults(WithVirtualizationSqlClientForm.queryResultsEmpty);
```
gashcrumb: What's this doin?

One thing that auto-form does is [pass along](https://github.com/syndesisio/syndesis-react/blob/master/app/ui-react/packages/auto-form/src/AutoForm.tsx#L136) all the [formik props](https://jaredpalmer.com/formik/docs/api/formik#props-1), among them are some imperative functions that let you adjust the form state.  This probably works because `validate` is called whenever there's state changes, but I'd be worried about breaking it if I go and refactor validation.  So then you could probably reset the form's state via [this function](https://jaredpalmer.com/formik/docs/api/formik#setvalues-fields-field-string-any-void) or maybe even just [reset the form](https://jaredpalmer.com/formik/docs/api/formik#resetform-nextvalues-values-void) perhaps?

mdrillin: yeah the intent was to clear the result table on change of state of one of the form controls.  I'll take another look

gashcrumb: Ah, I see, so it's not a case of resetting the actual form data.

Looking at the formik docs I don't really see a better way to do it either actually as it doesn't appear to have an `onChange` prop that we could wire up for this purpose.  So I'd say let's leave it as it is/

Might be good to leave a comment kinda mentioning what's going on so that someone doesn't come along later and decide to do something with `values` there and cause themselves a big re-rendering problem :smile:

mdrillin: Ok, I didn't see a better way either.  Added a comment and amended the PR...

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2406/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #454, opened 2019-06-11
fix: codemirror bugs

riccardo-forina:
* [x] codemirror cursor position jumping to the end
* [x] dropzone preventing switching template type (don't think we can address this without changing the UI, see [issue here](https://github.com/react-dropzone/react-dropzone/issues/826))
* [x] missing syntax highlighting

Fixes https://github.com/syndesisio/syndesis/issues/5649
Fixes https://github.com/syndesisio/syndesis/issues/5651

## Demo

codemirror


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2461/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

app/ui-react/packages/ui/src/Shared/TextEditor/velocity-lint.ts
```diff
@@ -1,12 +1,15 @@
...
 import { TemplateSymbol } from './template-symbol';
 
+import 'codemirror/mode/velocity/velocity.js';
+
+export type LintVelocity = 'velocity';
+
 export class VelocityLint extends AbstractLanguageLint {
   constructor() {
-    super(TemplateType.Velocity);
+    super('velocity');
```
gashcrumb: `LintVelocity` here?

app/ui-react/packages/ui/src/Shared/TextEditor/mustache-mode-lint.ts
```diff
@@ -1,12 +1,13 @@
...
+import * as Mustache from 'mustache';
 import { AbstractLanguageLint } from './abstract-language-lint';
 import { TemplateSymbol } from './template-symbol';
 
+export type LintMustache = 'mustache';
+
 export class MustacheModeLint extends AbstractLanguageLint {
   constructor() {
-    super(TemplateType.Mustache);
+    super('mustache');
```
gashcrumb: Oh, maybe make a const for this?

app/ui-react/packages/ui/src/Shared/TextEditor/velocity-lint.ts
gashcrumb: Sorry, missed that it was a type, why not use a const or enum here?

riccardo-forina: can't, it's a type :(

app/ui-react/packages/ui/src/Integration/Editor/template/TemplateStepTypeSelector.tsx
```diff
@@ -1,11 +1,12 @@
...
+  LintMustache,
+  LintVelocity,
+} from '../../../Shared/TextEditor';
 
-export enum TemplateType {
-  Freemarker = 'freemarker',
-  Mustache = 'mustache',
-  Velocity = 'velocity',
-}
+export type TemplateType = LintFreemarker | LintMustache | LintVelocity;
```
gashcrumb: I dunno, the problem then is we've special strings, I kinda thought the enums were a little more helpful.

riccardo-forina: The problem with the enum is that then the linter class is coupled with the template step mode selector. If we need a second specialized editor with different modes we'd need a second enum, but then which one should the linter class use?

This way it's a bit less type-safe because we are just enriching Codemirror modes with new modes, and delegate the control of the proper mode selection to whatever systems will be used to do that (radios for the templater)

gashcrumb: yeah, good point, also I don't feel strongly enough to worry about it too much either :smile:

Issue #397, opened 2019-06-03
🛠 API Designer button not set to blue to continue


gaughan:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x ] Bug report  
[ ] Documentation issue or request
</code></pre>

Workflow is complete in API Designer, but the button is not blue to proceed.
![image](https://user-images.githubusercontent.com/674767/58827740-bcb42780-8611-11e9-8e83-f8f6ed01b1f7.png)

PR #456, opened 2019-06-12
fix: api provider bugs

riccardo-forina:
Partially fixes https://github.com/syndesisio/syndesis/issues/5697
Partially fixes https://github.com/syndesisio/syndesis/issues/5600
Fixes https://github.com/syndesisio/syndesis/issues/5675
Fixes https://github.com/syndesisio/syndesis-react/issues/397

Since the spec is returned by an API, every time you go to the API definition you get the one relative to the last saved version of the integration. 
So if edit the definition, change your mind and try to edit it again, you'll lose the previous changes.

Also, this will completely delete whatever was present on the integration before the change in the spec. @zregvart can you confirm this is right?

### DIFF
--- a/app/ui-react/packages/api/src/useApiProviderSummary.tsx
+++ b/app/ui-react/packages/api/src/useApiProviderSummary.tsx
@@ -26,7 +26,7 @@ export function useApiProviderSummary(specification: string) {
           url: `${apiContext.apiUri}/apis/info`,
         });
         const summary = await response.json();
-        if (summary.errorCode || summary.errors) {
+        if (summary.errorCode && summary.errors) {
           throw new Error(
             summary.userMsg ||
               (summary.errors || [])
--- a/app/ui-react/packages/ui/src/Integration/Editor/IntegrationEditorLayout.tsx
+++ b/app/ui-react/packages/ui/src/Integration/Editor/IntegrationEditorLayout.tsx
@@ -121,6 +121,7 @@ export const IntegrationEditorLayout: React.FunctionComponent<
                     onClick={onSave}
                     href={saveHref}
                     disabled={isSaveLoading || isSaveDisabled}
+                    as={publishHref || onPublish ? 'default' : 'primary'}
                   >
                     {isSaveLoading ? (
                       <Loader size={'xs'} inline={true} />
--- a/app/ui-react/packages/ui/src/Integration/OperationsDropdown.css
+++ b/app/ui-react/packages/ui/src/Integration/OperationsDropdown.css
@@ -1,0 +1,4 @@
+.operations-dropdown .pf-c-dropdown__menu {
+  max-height: 300px;
+  overflow-y: auto;
+}
--- a/app/ui-react/packages/ui/src/Integration/OperationsDropdown.tsx
+++ b/app/ui-react/packages/ui/src/Integration/OperationsDropdown.tsx
@@ -1,5 +1,6 @@
 import { Dropdown, DropdownToggle } from '@patternfly/react-core';
 import * as React from 'react';
+import './OperationsDropdown.css';
 
 export interface IOperationsDropdownProps {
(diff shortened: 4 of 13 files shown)


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2472/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2490/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

zregvart:
> So if edit the definition, change your mind and try to edit it again, you'll lose the previous changes.

I think there's an explicit Save button, so the change your mind bit should be clicking on Cancel, when put this way I think this makes sense.

> Also, this will completely delete whatever was present on the integration before the change in the spec. @zregvart can you confirm this is right?

I think in the Angular UI we saved the integration before editing the API definition in Apicurio in that case changes to the flows will be preserved.

zregvart:
We could do a multipart `PUT /api/v1/integrations/{id}/specification`, one part Integration JSON another part the new OpenAPI JSON. That should fix it, right?

riccardo-forina:
Think so but why the id? This could happen for new integrations yet to be saved.

zregvart:
> Think so but why the id? This could happen for new integrations yet to be saved.

That's because we start with the OpenAPI document and create an Integration from it via `POST /api/v1/apis/generator` that returns the generated Integration JSON. You can't start from both Integration and OpenAPI document...

riccardo-forina:
Right but again, the integration could not be saved. Anyway, if we store the spec in the metadata we don’t need the specification endpoint anymore, don’t we? We just need the generator one, and that’s the one that should accept both an integration and the spec in the body

zregvart:
Sure we can do that, there's one caveat `POST /api/v1/apis/generator` is not without any side effects, we're storing the submitted OpenAPI document, the idea is that we serve from the running integration the same OpenAPI document the client provided.

I'm a bit hesitant to store the OpenAPI specification in metadata as it will be persisted alongside the Integration, and that will be (in general) a large JSON in JSON document which will lead to performance issues.

(we're kinda using the word `specification` wrong here, it means the OpenAPI document)

I'm thinking we probably need the specification endpoint, but we can create one without side effects. To refactor this fully we would:

 1. rename `POST /api/v1/apis/generator` to `POST /api/v1/integrations/specification` to generate integration from specification (receives multipart `specification`)
 1. rename `POST /api/v1/apis/info` to `POST /api/v1/integrations/specification/info` to generate integration summary from specification (receives multipart `specification`)
 1. add `PUT /api/v1/integrations/specification` to update the integration from specification (receives multipart `specification` and `integration`)
 1. add `POST /api/v1/integrations` to create integration and specification (receives multipart `specification` and `integration`)
 1. remove `PUT /api/v1/integrations/{id}/specification`

Does that make sense?

riccardo-forina:
Not sure about 4, doesn't that exist already? It's also a bit problematic to keep hold of the specification for us if it's not stored in the integration itself, the whole editor app is stateless - data gets passed from route to route - and we would have to add the specification string to this passing. It's definitely not ideal.

zregvart:
I think this discussion just uncovered a bunch of stuff we need discuss and design better, perhaps best done interactively.

I suggest that for the time being we can do the simplest thing, does just adding one endpoint, something like `PUT /api/v1/apis/generator` that takes both the integration and the OpenAPI document and returns the updated integration based on the OpenAPI document help? It can be invoked as soon as the user saves in Apicurio and the resulting integration can replace the one in the UI?

riccardo-forina:
I think that would patch this, yep!

zregvart:
I've created https://github.com/syndesisio/syndesis/issues/5719 and I'll work on it right away.

PR #458, opened 2019-06-12
refactor(oauth): prevent credentials POST for non-OAuth

zregvart:
When `GET /api/v1/connectors/{id}/credentials` returns `{}` we should not invoke a `POST /api/v1/connectors/{id}/credentials`.

This makes it so that the invocation of `useOAuthFlow` is conditional on the result of `useConnectorCredentials`.

### DIFF
--- a/app/ui-react/syndesis/src/modules/connections/pages/create/ConfigurationPage.tsx
+++ b/app/ui-react/syndesis/src/modules/connections/pages/create/ConfigurationPage.tsx
@@ -30,6 +30,43 @@ export interface IConfigurationPageRouteState {
   connector: Connector;
 }
 
+interface IOAuthFlowProps {
+  connectorId: string;
+  connectorName: string;
+  onSuccess: () => void;
+}
+
+const OAuthFlow: React.FunctionComponent<IOAuthFlowProps> = ({
+  connectorId,
+  connectorName,
+  onSuccess,
+}) => {
+  const { t } = useTranslation(['connections', 'shared']);
+
+  const { connectOAuth, isConnecting } = useOAuthFlow(
+    connectorId,
+    connectorName,
+    onSuccess
+  );
+
+  return (
+    <ConnectorAuthorization
+      i18nTitle={t('connections:create:configure:configurationTitle', {
+        name: connectorName,
+      })}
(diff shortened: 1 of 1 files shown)


app/ui-react/syndesis/src/modules/connections/pages/create/ConfigurationPage.tsx
```diff
@@ -124,24 +152,9 @@ export const ConfigurationPage: React.FunctionComponent = () => {
...
-                          'connections:create:configure:configurationButton',
-                          { name: connector.name }
-                        )}
-                        isConnecting={isConnecting}
-                        onConnect={connectOAuth}
-                      />
+                      <OAuthFlow connector={connector} history={history} />
                     ) : (
-                      <WithConnectorForm connector={connector} onSave={onSave}>
+                      <WithConnectorForm connector={connector} onSave={onSaveTFunction}>
```
riccardo-forina: fat fingered?

app/ui-react/syndesis/src/modules/connections/pages/create/ConfigurationPage.tsx
```diff
@@ -30,6 +30,46 @@ export interface IConfigurationPageRouteState {
   connector: Connector;
 }
 
+interface IOAuthFlowProps {
+  connector: Connector;
```
riccardo-forina: If we want to nitpick about this, I'd say that we could avoid passing the whole connector object and ask for the id and name instead, and instead of history we could ask for the callback to fire on a successful connect. And then this would be ready to be used in the creation flow too :)

app/ui-react/syndesis/src/modules/connections/pages/create/ConfigurationPage.tsx
zregvart: Totally, and I did fix that, I wonder how I commited the unfixed version though... thanks for finding this :)

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2502/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #463, opened 2019-06-13
fix: properly show the initial value for the form fields in all cases

riccardo-forina:
We should compute the `configuredProperties` in the configuration page since it depends on the mode (adding vs editing) and the page we are currently on.

Relative issue https://github.com/syndesisio/syndesis/issues/5686


riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2521/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

PR #472, opened 2019-06-13
fix: undo really trivial styling issue that happened, nothing to see here honest

gashcrumb:
has nothing to do with https://github.com/syndesisio/syndesis/issues/5732

@seanforyou23 shhh

### DIFF
--- a/app/ui-react/syndesis/src/index.tsx
+++ b/app/ui-react/syndesis/src/index.tsx
@@ -3,6 +3,9 @@ import {
   ServerEventsContext,
   WithServerEvents,
 } from '@syndesis/api';
+// Don't move this line, otherwise we anger the pf4
+import { App, IAppRoute, IAppRouteWithChildrens, WithConfig } from './app';
+// tslint:disable-next-line:ordered-imports
 import { createBrowserHistory } from '@syndesis/history';
 import { UnrecoverableError } from '@syndesis/ui';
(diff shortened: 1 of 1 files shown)


app/ui-react/syndesis/src/index.tsx
```diff
@@ -3,14 +3,16 @@ import {
   ServerEventsContext,
   WithServerEvents,
 } from '@syndesis/api';
+// Don't move this line, otherwise we anger the pf4
```
seanforyou23: I should have added a warning like this the first time around, good call!

riccardo-forina:
Wha? I don’t wanna know

seanforyou23:
@riccardo-forina it looks like if we import anything from `@syndesis/ui` before we import the main App `import { App, ... } from './app';` then the order of injected css styles is wrong in the resulting DOM. App needs to load first before other component stylesheets since it is what's importing the base styles (which contain resets).

Could probably refactor how these imports are structured but for now this get's us back to the stable point we've been seeing all week. I know you said you didn't wanna know, but I wanted you to know :)

riccardo-forina:
Now I want to unknown it all.

riccardo-forina:
PR Storybook available [here](https://circleci.com/api/v1.1/project/github/syndesisio/syndesis-react/2586/artifacts/0/home/circleci/repo/app/ui-react/doc/index.html)

